Issue #201, opened 2021-05-01
Embedded Orbit contexts (and possibly /review) don’t load when uBlock Origin is enabled


andymatuschak:
I've gotten a few reports from users that when uBlock Origin is installed, the embedded prompts are empty. And I've gotten one report that `/review` also fails to load! I haven't verified either, and I'm not sure why the latter would be. The former is perhaps unsurprising, since embedded iframe contexts are often quite locked down.

We should understand what's happening here and either adapt what we're doing to the constrained environment or at least warn the user.

qubist:
I'm not sure if this is relevant, but https://withorbit.com doesn't load when I have uBlock Origin enabled. (If there's a better place for this report I'll move or delete it)

**Browser:** Firefox 88.0 (64-bit)
**OS:** macOS 11.0.1 (20B29)

andymatuschak:
Yep, that's probably related, thanks!

PR #203, opened 2021-05-05
App fails to load when trackers are blocked

Talor-A:
Hey @andymatuschak , I was looking forward to trying out Orbit, but can't even access the site at all on Firefox with uBlock origin, since the Sentry script gets blocked. then `initializeReporter` throws, preventing `<App/>` from displaying at all because there is no Sentry global. 

Screen Shot 2021-05-05 at 9 43 14 AM


andymatuschak:
Outstanding—I was hoping it was something as easy as this!

Closes #201.

Issue #184, opened 2021-02-26
Project: add validation for core data types (API, storage)


andymatuschak:
I've played fast and loose with validation in Orbit so far, but particularly as we use on-device storage and open up a public API, we need to get serious about this. We need validators for all our core data types. When both client and server receive data from the network, they must validate that request/response. When reading data from disk, we should validate that data.

Ideally, our validation strategy is fairly automatic: either we're automatically generating code based on our TS types, or we're using some strategy to infer a validator at runtime (e.g. like [io-ts](https://github.com/gcanti/io-ts)).

I'd love help evaluating the approaches we might take to this, making a call, implementing the infrastructure, and adopting it appropriately throughout the system.

kirkbyo:
Gotcha, thanks for shining some light on the additional scope of this problem. I was overlooking the canonicalization routines. In that case, I am currently thinking of breaking down the tasks into the following PR's:

1. Set up validation at the HTTP layer on both the client and the backend. Since the `api` types consume a lot of the `core` types, this will be a little tricky. For this first PR, the API surface will be fully defined using JSON schemas however the definition of the core types will be duplicated. I won't convert the core types to be JSON schema's just yet to limit the blast radius. From here, I will set up `ajv` and `json-schema-to-typescript` and add a Github action to ensure that the Codegen has run before any merges. I hope this will expose any initial "gotcha's and allow us to pivot before investing too much.
2. Convert the `core` types to be JSON schema's instead and replace the canonicalization with `ajv`. I will remove the duplication introduced in the first PR, and instead reference the `core` types from the `API` schemas. 
3. Validate local serialization? I'll admit, I am not quite sure what you mean by "local serializations". Regardless, we can loop back at this point to see where else validation is needed.

Let me know if you have any other ideas or reservations about this approach!

andymatuschak:
Got it, sounds reasonable!

Regarding the local serialization: I mean loading from a local LevelDB database (a few implementations in `app`). But actually, it probably makes sense to hold off on that part, since I'm going to be substantially rearchitecting it in the coming weeks (or so I keep saying!).

PR #207, opened 2021-05-18
Upgrade Typescript to 4.2.4

kirkbyo:
To use `ajv` for the upcoming https://github.com/andymatuschak/orbit/issues/184 PR, the latest version of typescript is required. I also went ahead and standardized the typescript version throughout all the packages.

In the process of this upgrade it looks like Typescript is having trouble coercing `ActionLog` and `PromptActionLog` as equivalent. Specifically, it looks like the [taskParameters of RepetitionActionLog](https://github.com/andymatuschak/orbit/blob/master/packages/core/src/types/actionLog.ts#L21) and [taskParameters of PromptRepetitionActionLog](https://github.com/andymatuschak/orbit/blob/master/packages/core/src/types/promptActionLog.ts#L37) is the property blocking this equivalency despite them being equivalent in theory. To circumvent this problem a force cast between the two types were introduced in the code.

**How I tested this?**
- `yarn build` and `yarn test` at the root
- `yarn run dev` for the backend to make sure the backend still compiled correctly
- `yarn sync` for the `note-sync` package to make sure its CLI was not affected.
- `yarn storybook` for the UI

### DIFF
--- a/packages/app/src/model/databaseManager.ts
+++ b/packages/app/src/model/databaseManager.ts
@@ -1,5 +1,6 @@
 import OrbitAPIClient from "@withorbit/api-client";
 import {
+  ActionLog,
   ActionLogID,
   applyActionLogToPromptState,
   getActionLogFromPromptActionLog,
@@ -94,7 +95,7 @@ export default class DatabaseManager {
     const promptLogEntries = await Promise.all(
       [...entries].map(async (log) => ({
         log,
-        id: await getIDForActionLog(log),
+        id: await getIDForActionLog(getActionLogFromPromptActionLog(log)),
       })),
     );
 
@@ -180,7 +181,10 @@ export default class DatabaseManager {
       }
     }
 
-    await this.actionLogStore.saveActionLogs(entries);
+    // NOTE: As of Typescript 4.2.4 PromptActionLog is not equivalent ActionLog. Looks to be compiler limitation
+    await this.actionLogStore.saveActionLogs(
+      entries as { log: ActionLog; id: ActionLogID }[],
+    );
 
     await this.promptStateStore.savePromptStates(
       updates.map(({ newPromptState, taskID }) => ({
--- a/packages/app/src/model/databaseManager.ts
+++ b/packages/app/src/model/databaseManager.ts
@@ -183,7 +183,10 @@ export default class DatabaseManager {
 
     // NOTE: As of Typescript 4.2.4 PromptActionLog is not equivalent ActionLog. Looks to be compiler limitation
     await this.actionLogStore.saveActionLogs(
-      entries as { log: ActionLog; id: ActionLogID }[],
+      entries.map(({ id, log }) => ({
+        id,
+        log: getActionLogFromPromptActionLog(log),
+      })),
     );
 
     await this.promptStateStore.savePromptStates(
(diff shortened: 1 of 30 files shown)


kirkbyo:
Okay this PR should be good to go now!

Let me know if you would like the approach revisited. Especially since this does not solve the root of the problem but as you mentioned, hopefully this solution will be short lived.

packages/app/src/model/databaseManager.ts
```diff
@@ -180,7 +181,10 @@ export default class DatabaseManager {
       }
     }
 
-    await this.actionLogStore.saveActionLogs(entries);
+    // NOTE: As of Typescript 4.2.4 PromptActionLog is not equivalent ActionLog. Looks to be compiler limitation
```
andymatuschak, line 184: If possible, I'd love to confine this narrowing issue to our bottleneck function (`getActionLogFromPromptActionLog`), so that if the conversion does indeed involve some actual value re-mapping, the call-sites will all get the new types as appropriate. So could we make this something like?:

```
    await this.actionLogStore.saveActionLogs(
      entries.map(({log, id}) => ({ log: getActionLogFromPromptActionLog(log); id })),
    );
```

packages/ui/src/components/Card.tsx
```diff
@@ -41,9 +43,14 @@ function getQAPromptContents<PT extends QAPromptTask | ApplicationPromptTask>(
...
-      return applicationReviewItem.prompt.variants[
-        applicationReviewItem.taskParameters.variantIndex
-      ];
+      const taskParameters = applicationReviewItem.taskParameters as {
+        variantIndex: number;
+      };
+      const applicationPrompt = applicationReviewItem.prompt as ApplicationPrompt;
+
+      return applicationPrompt.variants[taskParameters.variantIndex];
+    case clozePromptType:
```
andymatuschak, line 52: Odd—you shouldn't have to do this, since the generic constraint should exclude clozes! I'm guessing that the compiler couldn't figure that out?

andymatuschak (changes requested):
I’m sorry you had to deal with this—broadly looks good! Would sleep better with this one spot funneled through our bottleneck, then we’re good to go.

packages/ui/src/components/Card.tsx
kirkbyo: Yep this unfortunate. It looks like through bumping the package the code below is giving typescript trouble across the board.
 
```typescript
export type PromptTaskParametersOf<
  PT extends PromptTask
> = PT extends AbstractPromptTask<any, any, infer PTP> ? PTP : never;
```

Typescript seems unable to coerce the type any longer (see screenshot) which prevents it from detecting that this case is unreachable.

Screen Shot 2021-05-18 at 2 09 31 PM

Without including that switch case the following error occurs:

```bash
$ tsc -b packages/tsconfig.json
packages/ui/src/components/Card.tsx:38:4 - error TS2366: Function lacks ending return statement and return type does not include 'undefined'.

38 ): QAPromptContents {
      ~~~~~~~~~~~~~~~~
```

Adding the cloze case resolves the issue.

I am worried this lack of type information will deteriorate type safety across the board. Any thoughts?

packages/app/src/model/databaseManager.ts
kirkbyo: I will make the change! 

I originally omitted using the bottle neck function since it seemed a bit wasteful from a performance standpoint to map through an array just to cast it for typescript. However, I guess the performance impact is likely negligible.

andymatuschak: Makes sense. If the JIT's doing its job, it may be inlined out! :)

PR #208, opened 2021-05-20
Add validation to the /actionLogs endpoint

kirkbyo:
Related: #184 

This PR looks to add validation to the `/actionLogs` endpoint as part of the ongoing effort to add validation across the app. To accomplish this, a JSON schema is generated based on a portion of the API spec (defined by `ValidatableSpec`). JSON schema was chosen since it extends well to the typescript environment and can create constraints that Typescript cannot (i.e. integer type constraint and much more!). This allows the endpoint to focus on the business logic instead of worrying about validation. Using this format will also allow for public facing documentation to be generated with ease.

## Testing Methodology
<details>
  <summary>Manuel Integration-ish tests (expand for error messages example)</summary>

#### [GET] `/actionLogs`
**it does not allow limit to be a negative integer**
Request: 
```
[GET] http://localhost:5001/metabook-system/us-central1/api/actionLogs?limit=-5
```

Response:
```
{
  "errors": [
    {
      "message": "query/properties/limit/minimum should be >= 1"
    }
  ]
}
```

**it does not allow limit to be a floating point**
Request:
```
[GET] http://localhost:5001/metabook-system/us-central1/api/actionLogs?limit=5.4
```
Response:
```
{
    "errors": [
        {
            "message": "query/properties/limit/type should be integer"
        }
    ]
}
```

**it does not allow createdAfterID to be a string**
Request:
```
[GET] http://localhost:5001/metabook-system/us-central1/api/actionLogs?createdAfterID=2
```

Response:
```
{
    "errors": [
        {
            "message": "query/properties/createdAfterID/type should be string"
        }
    ]
}
```

**it succeeds**
Request:
```
[GET] http://localhost:5001/metabook-system/us-central1/api/actionLogs?createdAfterID="someId"&limit=1000
```

Response:
```
401 Unauthorized  (validation middleware completed without any issues)
```

#### [PATCH] `/actionLogs`

**it fails when extra properties are provided**
Request:
```
[PATCH] http://localhost:5001/metabook-system/us-central1/api/actionLogs
[{
    "id": "someId",
    "data": {
        "timestampMillis": 15,
        "taskID": "12",
        "actionLogType": "ingest",
        "provenance": null
    },
    "someExtraProperty": true
}]
```
Response:
```
{
    "errors": [
        {
            "message": "body/items/additionalProperties should NOT have additional properties"
        }
    ]
}
```

**it fails when actionLogType is invalid**
Request:
```
[PATCH] http://localhost:5001/metabook-system/us-central1/api/actionLogs
[{
    "id": "someId",
    "data": {
        "timestampMillis": 15,
        "taskID": "12",
        "actionLogType": "invalid_action_log_type",
        "provenance": null
    }
}]
```
Response:
```
{
    "errors": [
        {
            "message": "body/items/properties/data/anyOf/0/properties/actionLogType/enum should be equal to one of the allowed values"
        },
        {
            "message": "body/items/properties/data/anyOf/1/additionalProperties should NOT have additional properties"
        },
        {
            "message": "body/items/properties/data/anyOf/2/additionalProperties should NOT have additional properties"
        },
        {
            "message": "body/items/properties/data/anyOf/3/additionalProperties should NOT have additional properties"
        },
        {
            "message": "body/items/properties/data/anyOf should match some schema in anyOf"
        }
    ]
}
```

**it succeeds**
Request:
```
[PATCH] http://localhost:5001/metabook-system/us-central1/api/actionLogs
[{
    "id": "someId",
    "data": {
        "timestampMillis": 15,
        "taskID": "12",
        "actionLogType": "ingest",
        "provenance": null
    }
}]
```
Response:
```
401 Unauthorized  (validation middleware completed without any issues)
```
</details>

Ran the iOS app to make sure there were no warnings being printed from the api-client.

## Tradeoffs
- Error messages are not great and can be confusing when dealing with union types. For example, see the case of the invalid `actionLogType` value. The ajv compiler detects that the type is not valid, however it also flags that the schema does not match any of the other possible definitions. Basically, its ability to narrow the type across union types is not great.
- This current solution might be tricky to extend to dynamic endpoints like `/attachments/:id`. Not sure how to handle this case at the moment.

### DIFF
--- a/packages/api-client/src/requestManager.ts
+++ b/packages/api-client/src/requestManager.ts
@@ -1,16 +1,19 @@
-import { API } from "@withorbit/api";
+import { API, APIValidator } from "@withorbit/api";
 import { APIConfig } from "./apiConfig";
 import * as Network from "./util/fetch";
 
 export class RequestManager<API extends API.Spec> {
   private readonly config: APIConfig;
   private readonly authorizeRequest?: () => Promise<AuthenticationConfig>;
+  private readonly validator: APIValidator;
 
   constructor(
     config: APIConfig,
+    validator: APIValidator,
     authorizeRequest?: () => Promise<AuthenticationConfig>,
   ) {
     this.config = config;
+    this.validator = validator;
     this.authorizeRequest = authorizeRequest;
   }
 
@@ -93,7 +96,18 @@ export class RequestManager<API extends API.Spec> {
         >;
       } else {
         const json = await response.json();
-        // TODO validate
+
+        const validationResult = this.validator.validateResponse(
+          { path, method, ...requestData },
+          json,
+        );
+        if (validationResult !== true) {
+          console.warn(
+            `Validation Error: (${method}) ${path} did not match the expected value for this route`,
+          );
+          console.warn(validationResult);
+        }
+
         return json as API.RouteResponseData<API[Path][Method]>;
       }
     } else {
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -6,10 +6,14 @@
   "types": "./dist/index.d.ts",
   "private": true,
   "scripts": {
-    "build": "tsc -b"
+    "build": "tsc -b",
+    "generateSchema": "typescript-json-schema src/orbitAPI.ts ValidatableSpec -o src/orbitAPISchema.json --noExtraProps --required --ignoreErrors --strictNullChecks"
   },
-  "dependencies": {},
+  "dependencies": {
+    "ajv": "8.4.0"
+  },
   "devDependencies": {
-    "typescript": "^4.2.4"
+    "typescript": "^4.2.4",
+    "typescript-json-schema": "^0.50.0"
   }
 }
--- a/packages/api/src/orbitAPI.ts
+++ b/packages/api/src/orbitAPI.ts
@@ -9,9 +9,10 @@ import {
   PromptTaskID,
 } from "@withorbit/core";
 import { BlobLike } from "./genericHTTPAPI";
+import { RequiredSpec } from "./util/requiredSpec";
 
 // Meant to conform to genericHTTPAPI/Spec, but I can't declare conformance without running into obscure Typescript limitations.
-export type Spec = {
+type SpecLegacy = {
   "/taskStates": {
     GET: {
       query:
@@ -27,21 +28,6 @@ export type Spec = {
           ))
         | { ids: PromptTaskID[] };
       response: ResponseList<"taskState", PromptTaskID, PromptState>;
-    };
-  };
-
-  "/actionLogs": {
-    GET: {
-      query: {
-        limit?: number;
-        createdAfterID?: ActionLogID;
-      };
-      response: ResponseList<"actionLog", ActionLogID, ActionLog>;
-    };
-
-    PATCH: {
-      body: { id: ActionLogID; data: ActionLog }[];
-      response: void;
     };
   };
 
@@ -88,6 +74,41 @@ export type Spec = {
     returns application/json encoded ResponseObject<"attachmentIDReference", AttachmentID, AttachmentIDReference>
    */
 };
+
+export type ValidatableSpec = {
+  "/actionLogs"?: {
+    GET?: {
+      query: {
+        /**
+         * @minimum 1
+         * @default 100
+         * @TJS-type integer
+         */
+        limit?: number;
+        /**
+         * Created After ActionLogID
+         * @TJS-type string
+         */
+        createdAfterID?: ActionLogID;
+      };
+      response?: ResponseList<"actionLog", ActionLogID, ActionLog>;
+    };
+
+    PATCH?: {
+      body: {
+        /**
+         * Created After ActionLogID
+         * @TJS-type string
+         */
+        id: ActionLogID;
+        data: ActionLog;
+      }[];
+      response?: null;
+    };
+  };
+};
+
+export type Spec = RequiredSpec<ValidatableSpec> & SpecLegacy;
 
 export type ResponseList<
   ObjectTypeString extends string,
--- a/packages/api/src/orbitAPI.ts
+++ b/packages/api/src/orbitAPI.ts
@@ -85,10 +85,6 @@ export type ValidatableSpec = {
          * @TJS-type integer
          */
         limit?: number;
-        /**
-         * Created After ActionLogID
-         * @TJS-type string
-         */
         createdAfterID?: ActionLogID;
       };
       response?: ResponseList<"actionLog", ActionLogID, ActionLog>;
@@ -96,10 +92,6 @@ export type ValidatableSpec = {
 
     PATCH?: {
       body: {
-        /**
-         * Created After ActionLogID
-         * @TJS-type string
-         */
         id: ActionLogID;
         data: ActionLog;
       }[];
@@ -126,6 +118,9 @@ export type ResponseObject<
   DataType
 > = {
   objectType: ObjectType;
+  /**
+   * @TJS-type string
+   */
   id: IDType;
   data: DataType;
 };
--- a/packages/api/src/orbitAPISchema.json
+++ b/packages/api/src/orbitAPISchema.json
@@ -2,6 +2,40 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "additionalProperties": false,
     "definitions": {
+        "IngestActionLog": {
+            "additionalProperties": false,
+            "properties": {
+                "actionLogType": {
+                    "enum": [
+                        "ingest"
+                    ],
+                    "type": "string"
+                },
+                "provenance": {
+                    "anyOf": [
+                        {
+                            "$ref": "#/definitions/TaskProvenance"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ]
+                },
+                "taskID": {
+                    "type": "string"
+                },
+                "timestampMillis": {
+                    "type": "number"
+                }
+            },
+            "required": [
+                "actionLogType",
+                "provenance",
+                "taskID",
+                "timestampMillis"
+            ],
+            "type": "object"
+        },
         "Partial<TaskMetadata>": {
             "additionalProperties": false,
             "properties": {
@@ -28,6 +62,105 @@
                 "web"
             ],
             "type": "string"
+        },
+        "RepetitionActionLog": {
+            "additionalProperties": false,
+            "properties": {
+                "actionLogType": {
+                    "enum": [
+                        "repetition"
+                    ],
+                    "type": "string"
+                },
+                "context": {
+                    "type": [
+                        "null",
+                        "string"
+                    ]
+                },
+                "outcome": {
+                    "type": "string"
+                },
+                "parentActionLogIDs": {
+                    "items": {
+                        "additionalProperties": false,
+                        "properties": {
+                        },
+                        "type": "object"
+                    },
+                    "type": "array"
+                },
+                "taskID": {
+                    "type": "string"
+                },
+                "taskParameters": {
+                    "anyOf": [
+                        {
+                            "additionalProperties": {
+                                "type": [
+                                    "null",
+                                    "string",
+                                    "number"
+                                ]
+                            },
+                            "type": "object"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ]
+                },
+                "timestampMillis": {
+                    "type": "number"
+                }
+            },
+            "required": [
+                "actionLogType",
+                "context",
+                "outcome",
+                "parentActionLogIDs",
+                "taskID",
+                "taskParameters",
+                "timestampMillis"
+            ],
+            "type": "object"
+        },
+        "RescheduleActionLog": {
+            "additionalProperties": false,
+            "properties": {
+                "actionLogType": {
+                    "enum": [
+                        "reschedule"
+                    ],
+                    "type": "string"
+                },
+                "newTimestampMillis": {
+                    "type": "number"
+                },
+                "parentActionLogIDs": {
+                    "items": {
+                        "additionalProperties": false,
+                        "properties": {
+                        },
+                        "type": "object"
+                    },
+                    "type": "array"
+                },
+                "taskID": {
+                    "type": "string"
+                },
+                "timestampMillis": {
+                    "type": "number"
+                }
+            },
+            "required": [
+                "actionLogType",
+                "newTimestampMillis",
+                "parentActionLogIDs",
+                "taskID",
+                "timestampMillis"
+            ],
+            "type": "object"
         },
         "TaskProvenance": {
             "additionalProperties": false,
@@ -65,6 +198,43 @@
                 "url"
             ],
             "type": "object"
+        },
+        "UpdateMetadataActionLog": {
+            "additionalProperties": false,
+            "properties": {
+                "actionLogType": {
+                    "enum": [
+                        "updateMetadata"
+                    ],
+                    "type": "string"
+                },
+                "parentActionLogIDs": {
+                    "items": {
+                        "additionalProperties": false,
+                        "properties": {
+                        },
+                        "type": "object"
+                    },
+                    "type": "array"
+                },
+                "taskID": {
+                    "type": "string"
+                },
+                "timestampMillis": {
+                    "type": "number"
+                },
+                "updates": {
+                    "$ref": "#/definitions/Partial<TaskMetadata>"
+                }
+            },
+            "required": [
+                "actionLogType",
+                "parentActionLogIDs",
+                "taskID",
+                "timestampMillis",
+                "updates"
+            ],
+            "type": "object"
         }
     },
     "properties": {
@@ -78,8 +248,10 @@
                             "additionalProperties": false,
                             "properties": {
                                 "createdAfterID": {
-                                    "description": "Created After ActionLogID",
-                                    "type": "string"
+                                    "additionalProperties": false,
+                                    "properties": {
+                                    },
+                                    "type": "object"
                                 },
                                 "limit": {
                                     "default": 100,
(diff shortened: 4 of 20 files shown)


packages/backend/src/api.ts
```diff
@@ -36,6 +42,7 @@ export function createAPIApp(): express.Application {
     next();
   });
   app.use(traceAPICall);
+  app.use(validateRequest(routeValidator));
```
kirkbyo, line 45: Any preference on having this as a middleware vs. a function wrapper like `authenticateRequest`? I figured this way validation happens across the board without having to think about it. My only concern is it is too much indirection and would look _magical_ to a new dev on the project?

packages/backend/src/api.ts
```diff
@@ -23,6 +24,11 @@ const traceAPICall: express.RequestHandler = (request, _, next) => {
   next();
 };
 
+const routeValidator = new OrbitAPIValidator({
+  allowUnsupportedRoute: true,
+  mutateWithDefaultValues: true,
```
kirkbyo, line 29: allows request queries and bodies to be _augmented_ with the default values defined in the JSON schema

packages/api/package.json
```diff
@@ -6,10 +6,14 @@
   "types": "./dist/index.d.ts",
   "private": true,
   "scripts": {
-    "build": "tsc -b"
+    "build": "tsc -b",
+    "generateSchema": "typescript-json-schema src/orbitAPI.ts ValidatableSpec -o src/orbitAPISchema.json --noExtraProps --required --ignoreErrors --strictNullChecks"
```
kirkbyo, line 10: `--ignoreErrors`: This is an unfortunate flag to use.

<details>
<summary> Without the flag this error is generated</summary>

```bash
orbit/node_modules/@types/react-native/globals.d.ts (49,13): Subsequent variable declarations must have the same type.  Variable 'Blob' must be of type '{ new (blobParts?: BlobPart[] | undefined, options?: BlobPropertyBag | undefined): Blob; prototype: Blob; }', but here has type '{ new (blobParts?: (string | Blob)[] | undefined, options?: BlobOptions | undefined): Blob; prototype: Blob; }'.
orbit/node_modules/@types/react-native/globals.d.ts (54,15): Duplicate identifier 'FormData'.
orbit/node_modules/@types/react-native/globals.d.ts (110,5): Subsequent property declarations must have the same type.  Property 'body' must be of type 'BodyInit | null | undefined', but here has type 'BodyInit_ | undefined'.
orbit/node_modules/@types/react-native/globals.d.ts (119,5): Subsequent property declarations must have the same type.  Property 'signal' must be of type 'AbortSignal | null | undefined', but here has type 'AbortSignal | undefined'.
orbit/node_modules/@types/react-native/globals.d.ts (137,14): Duplicate identifier 'RequestInfo'.
orbit/node_modules/@types/react-native/globals.d.ts (156,13): Subsequent variable declarations must have the same type.  Variable 'Response' must be of type '{ new (body?: BodyInit | null | undefined, init?: ResponseInit | undefined): Response; prototype: Response; error(): Response; redirect(url: string, status?: number | undefined): Response; }', but here has type '{ new (body?: BodyInit_ | undefined, init?: ResponseInit | undefined): Response; prototype: Response; error: () => Response; redirect: (url: string, status?: number | undefined) => Response; }'.
orbit/node_modules/@types/react-native/globals.d.ts (233,5): Subsequent property declarations must have the same type.  Property 'abort' must be of type 'ProgressEvent<XMLHttpRequestEventTarget>', but here has type 'ProgressEvent<EventTarget>'.
orbit/node_modules/@types/react-native/globals.d.ts (234,5): Subsequent property declarations must have the same type.  Property 'error' must be of type 'ProgressEvent<XMLHttpRequestEventTarget>', but here has type 'ProgressEvent<EventTarget>'.
orbit/node_modules/@types/react-native/globals.d.ts (235,5): Subsequent property declarations must have the same type.  Property 'load' must be of type 'ProgressEvent<XMLHttpRequestEventTarget>', but here has type 'ProgressEvent<EventTarget>'.
orbit/node_modules/@types/react-native/globals.d.ts (236,5): Subsequent property declarations must have the same type.  Property 'loadend' must be of type 'ProgressEvent<XMLHttpRequestEventTarget>', but here has type 'ProgressEvent<EventTarget>'.
orbit/node_modules/@types/react-native/globals.d.ts (237,5): Subsequent property declarations must have the same type.  Property 'loadstart' must be of type 'ProgressEvent<XMLHttpRequestEventTarget>', but here has type 'ProgressEvent<EventTarget>'.
orbit/node_modules/@types/react-native/globals.d.ts (238,5): Subsequent property declarations must have the same type.  Property 'progress' must be of type 'ProgressEvent<XMLHttpRequestEventTarget>', but here has type 'ProgressEvent<EventTarget>'.
orbit/node_modules/@types/react-native/globals.d.ts (239,5): Subsequent property declarations must have the same type.  Property 'timeout' must be of type 'ProgressEvent<XMLHttpRequestEventTarget>', but here has type 'ProgressEvent<EventTarget>'.
orbit/node_modules/@types/react-native/globals.d.ts (280,14): Duplicate identifier 'XMLHttpRequestResponseType'.
orbit/node_modules/@types/react-native/globals.d.ts (287,15): Duplicate identifier 'URL'.
orbit/node_modules/@types/react-native/globals.d.ts (312,15): Duplicate identifier 'URLSearchParams'.
orbit/node_modules/@types/react-native/globals.d.ts (346,5): Subsequent property declarations must have the same type.  Property 'onopen' must be of type '((this: WebSocket, ev: Event) => any) | null', but here has type '(() => void) | null'.
orbit/node_modules/@types/react-native/globals.d.ts (347,5): Subsequent property declarations must have the same type.  Property 'onmessage' must be of type '((this: WebSocket, ev: MessageEvent<any>) => any) | null', but here has type '((event: WebSocketMessageEvent) => void) | null'.
orbit/node_modules/@types/react-native/globals.d.ts (348,5): Subsequent property declarations must have the same type.  Property 'onerror' must be of type '((this: WebSocket, ev: Event) => any) | null', but here has type '((event: WebSocketErrorEvent) => void) | null'.
orbit/node_modules/@types/react-native/globals.d.ts (349,5): Subsequent property declarations must have the same type.  Property 'onclose' must be of type '((this: WebSocket, ev: CloseEvent) => any) | null', but here has type '((event: WebSocketCloseEvent) => void) | null'.
orbit/node_modules/@types/react-native/globals.d.ts (352,13): Subsequent variable declarations must have the same type.  Variable 'WebSocket' must be of type '{ new (url: string, protocols?: string | string[] | undefined): WebSocket; prototype: WebSocket; readonly CLOSED: number; readonly CLOSING: number; readonly CONNECTING: number; readonly OPEN: number; }', but here has type '{ new (uri: string, protocols?: string | string[] | null | undefined, options?: { [optionName: string]: any; headers: { [headerName: string]: string; }; } | null | undefined): WebSocket; prototype: WebSocket; readonly CLOSED: number; readonly CLOSING: number; readonly CONNECTING: number; readonly OPEN: number; }'.
orbit/node_modules/@types/react-native/globals.d.ts (376,15): Duplicate identifier 'AbortSignal'.
orbit/node_modules/@types/react-native/globals.d.ts (389,15): Duplicate identifier 'AbortController'.
orbit/node_modules/@types/react-native/globals.d.ts (414,14): Subsequent property declarations must have the same type.  Property 'error' must be of type 'DOMException | null', but here has type 'Error | null'.
orbit/node_modules/@types/react-native/globals.d.ts (422,14): Subsequent property declarations must have the same type.  Property 'result' must be of type 'string | ArrayBuffer | null', but here has type 'string | ArrayBuffer'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (2008,11): Duplicate identifier 'AbortController'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (2019,13): Duplicate identifier 'AbortController'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (2029,11): Duplicate identifier 'AbortSignal'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (2041,13): Duplicate identifier 'AbortSignal'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (5620,11): Duplicate identifier 'FormData'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (5630,13): Duplicate identifier 'FormData'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (16046,11): Duplicate identifier 'URL'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (16063,13): Duplicate identifier 'URL'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (16073,11): Duplicate identifier 'URLSearchParams'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (16106,13): Duplicate identifier 'URLSearchParams'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (20030,6): Duplicate identifier 'RequestInfo'.
orbit/node_modules/typescript/lib/lib.dom.d.ts (20230,6): Duplicate identifier 'XMLHttpRequestResponseType'.
(node:20182) UnhandledPromiseRejectionWarning: Error: No output definition. Probably caused by errors prior to this?
    at Object.<anonymous> (orbit/node_modules/typescript-json-schema/dist/typescript-json-schema.js:1290:23)
    at step (orbit/node_modules/typescript-json-schema/dist/typescript-json-schema.js:44:23)
    at Object.next (orbit/node_modules/typescript-json-schema/dist/typescript-json-schema.js:25:53)
    at orbit/node_modules/typescript-json-schema/dist/typescript-json-schema.js:19:71
    at new Promise (<anonymous>)
    at __awaiter (orbit/node_modules/typescript-json-schema/dist/typescript-json-schema.js:15:12)
    at Object.exec (orbit/node_modules/typescript-json-schema/dist/typescript-json-schema.js:1269:12)
    at Object.run (orbit/node_modules/typescript-json-schema/dist/typescript-json-schema-cli.js:53:30)
    at Object.<anonymous> (orbit/node_modules/typescript-json-schema/bin/typescript-json-schema:3:5)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:20182) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:20182) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
</details>

I did quite a bit of research to try and resolve these errors but the cause of them is unclear to me since we are not even importing/using most of the types described in the message. Regardless, the schema is generated correctly.

andymatuschak, line 10: Yikes. This is… probably fine, assuming we're generally going to be running a "real" build simultaneously.

packages/api/src/orbitAPI.ts
```diff
@@ -89,6 +75,41 @@ export type Spec = {
...
+      query: {
+        /**
+         * @minimum 1
+         * @default 100
+         * @TJS-type integer
+         */
+        limit?: number;
+        /**
+         * Created After ActionLogID
+         * @TJS-type string
```
andymatuschak, line 90: Idle wondering: is it possible to add this annotation to the definition of `ActionLogID` so that it propagates to all client sites, rather than needing to repeat at each one?

packages/api/src/util/requiredSpec.ts
```diff
@@ -0,0 +1,31 @@
...
+ * type Spec = RequiredSpec<{
+ *  "/hello": {
+ *    GET: {
+ *      query: { name: string };
+ *      response: string;
+ *    }
+ *  }
+ * }>
+ * ```
+ */
```
andymatuschak, line 22: Appreciate the nice example in the comment.

packages/api/src/orbitAPISchema.json
```diff
@@ -0,0 +1,527 @@
...
+                },
+                "PATCH": {
+                    "additionalProperties": false,
+                    "properties": {
+                        "body": {
+                            "items": {
+                                "additionalProperties": false,
+                                "properties": {
+                                    "data": {
+                                        "anyOf": [
```
andymatuschak, line 326: It's a shame that the definition for `ActionLog` ends up repeated here. It's probably not that big a deal, but if we do end up validating API responses in `app`, we'll be shipping this JSON file to browsers. I don't suppose there's a flag we're missing that can indirect types? If not I imagine we may skip validating responses client-side.

packages/api/src/validation/mockAPIValiation.ts
```diff
@@ -0,0 +1,11 @@
+import { APIValidator } from "./apiValidator";
+
+export class MockOrbitAPIValidation implements APIValidator {
```
andymatuschak, line 3: If this is probably only going to be used in `api-client`'s tests, it may be better to move this into that package to avoid exporting it as public interface from this module.

packages/backend/src/api.ts
andymatuschak, line 45: I think this is fine for now! The main reason I made `authenticateRequest` a function wrapper was that it needed to _output_ the authenticated user ID, and I couldn't figure out a typesafe way to do that with Express middleware normally. (But your question reflects a good instinct, I think)

packages/backend/src/util/validateRequest.ts
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  if (validationResult === true) {
+    next();
+  } else {
+    res.status(400).send(validationResult);
+  }
+};
+
+function convertNumbersToPrimitiveType(object: Record<string, unknown>) {
```
andymatuschak, line 27: Daring! And slightly magic: could you add a bit of documentation describing what's going on here? I worry that in the future we may add some string query parameter to some API for which a client may _sometimes_ specify a number… and then the type for the same field on the same route may vary at runtime! (I don't suppose we can straightforwardly read the "intended" type via ajv at runtime?)

packages/api-client/src/requestManager.ts
```diff
@@ -93,7 +96,18 @@ export class RequestManager<API extends API.Spec> {
...
+        const validationResult = this.validator.validateResponse(
+          { path, method, ...requestData },
+          json,
+        );
+        if (validationResult !== true) {
+          console.warn(
+            `Validation Error: (${method}) ${path} did not match the expected value for this route`,
+          );
+          console.warn(validationResult);
+        }
```
andymatuschak, line 109: Good question! One easy way: run `yarn dev` in `packages/web-component` then navigate to `http://localhost:5000` to find a web page with some test embedded prompts.

packages/api/src/validation/orbitAPIValidator.ts
```diff
@@ -0,0 +1,87 @@
...
+  APIValidatorError,
+  APIValidator,
+} from "./apiValidator";
+
+type OrbitAPIValidatorConfig = {
+  mutateWithDefaultValues: boolean;
+  allowUnsupportedRoute: boolean;
+};
+
+export class OrbitAPIValidator implements APIValidator {
```
andymatuschak, line 18: I'd love to see some tests on this when you get a chance! I have trouble being fully confident in e.g. the implementation of `allowUnsupportedRoute`, since it depends on details of ajv's validation output.

andymatuschak (approved):
This is looking really great, Ozzie! I like the general strategy. A couple small inline questions / comments; nothing blocking. Feel free to address whichever ones seem appropriate and merge at your leisure.

packages/api/src/orbitAPI.ts
kirkbyo: Good idea, I actually did not consider this beforehand. Unfortunately, it looks `typescript-json-schema` is not cooperating. In theory, what you mentioned should work, however `ActionLogID` is defined as a `type` (rightly so), but it looks like the library does not use the comment for it. I also tried creating a _box_ of sorts:

```typescript
interface BoxActionLogID {
  id: string & { __actionLogIDOpaqueType: never };
}
export type ActionLogID = BoxActionLogID["id"];
```

This works! But, only when the id is not defined as an array, so still not great. I believe this to be a bug. If I have sometime soon, I'll see if I can create an issue on their end.

packages/api/src/orbitAPISchema.json
kirkbyo: Ah yes, I noticed this as well. For this to work, we could enable the `aliasRefs` flag. From their docs:
```
  --aliasRefs           Create shared ref definitions for the type aliases.          [boolean] [default: false]
```
However, this flag seems to create a lot of noise since every type will now be defined and it appears to be broken for generics. Instead, I converted the `ActionLog` to be defined using `interface`s rather than `type` which indicates to export the type as a shared ref. This ends up reducing the lint count quite a bit!

Example:
```json
"data": {
  "anyOf": [
      {
          "$ref": "#/definitions/IngestActionLog"
      },
      {
          "$ref": "#/definitions/RepetitionActionLog"
      },
      {
          "$ref": "#/definitions/RescheduleActionLog"
      },
      {
          "$ref": "#/definitions/UpdateMetadataActionLog"
      }
  ]
},
```

packages/api/src/validation/orbitAPIValidator.ts
kirkbyo: Good call, on my list! I tried setting up jest for api package earlier but I was getting some weird jest/build errors and was planning on looping back 🙂 Thanks for the reminder!

packages/backend/src/util/validateRequest.ts
kirkbyo: I wasn't too keen on this code either. Your comment got me thinking, why doesn't `ajv` just do this inference? They already know which type it is supposed to be! Turns out, [they do](https://ajv.js.org/coercion.html), oops. I enabled the flag and got rid of this code.

kirkbyo:
Going to go ahead and marge this now. I will add testing in a follow-up PR since it required a little refactor. 

Let me know if you have any comments on any of the changes. Happy to address them in a follow up PR!

packages/backend/src/util/validateRequest.ts
andymatuschak: Wow, fantastic!!

packages/api/src/orbitAPISchema.json
andymatuschak: Hilariously, I remember switching `ActionLog` from an `interface` to a `type` a while back because TypeScript couldn't infer something I expected it to. Alas, I don't remember the specifics. I think the issue is that `interface`s are "open"—i.e. subtypes can add fields—and this broke some exhaustiveness guarantee I needed. Assuming that the whole project still builds, this should be fine!

PR #209, opened 2021-05-22
Add tests to OrbitAPIValidator

kirkbyo:
Add a few test cases to the OrbitAPIValidator introduced in https://github.com/andymatuschak/orbit/pull/208. It required setting up `jest` for the `api` package.

### DIFF
--- a/packages/api/jest.config.cjs
+++ b/packages/api/jest.config.cjs
@@ -1,0 +1,15 @@
+module.exports = {
+  testEnvironment: "node",
+  testMatch: ["**/?(*.)+(spec|test).ts?(x)"],
+  testPathIgnorePatterns: ["dist", "node_modules"],
+  transformIgnorePatterns: [],
+  transform: {
+    "\\.[jt]s$": ["babel-jest", { cwd: __dirname }],
+  },
+  // Jest struggles to read these mappings from the multiformats packages, which seem to be specifying them just fine as far as I can tell.
+  moduleNameMapper: {
+    "^multiformats$": "multiformats/cjs/src/index.js",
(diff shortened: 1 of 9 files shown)


packages/api/jest.config.cjs
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  testEnvironment: "node",
+  testMatch: ["**/?(*.)+(spec|test).ts?(x)"],
+  testPathIgnorePatterns: ["dist", "node_modules"],
+  transformIgnorePatterns: [],
+  transform: {
+    "\\.[jt]s$": ["babel-jest", { cwd: __dirname }],
+  },
+  // Jest struggles to read these mappings from the multiformats packages, which seem to be specifying them just fine as far as I can tell.
+  moduleNameMapper: {
```
andymatuschak, line 10: :roll_eyes: Such a shame we have to repeat this in all our packages' configs.

andymatuschak (approved):
Looks great. Could I trouble you (either in this PR or a new one) for a test exercising the type coercion semantics we expect?

packages/api/jest.config.cjs
kirkbyo: Is there any particular reason why we can't set this up at the packages root? Is the difference in babel configs the blocker?

kirkbyo:
Done, good call!

packages/api/jest.config.cjs
andymatuschak: Maybe we can! I'm not sure how configuration "inheritance" works for Jest.

PR #210, opened 2021-05-23
Simplify ActionLogID annotations and add an action to make sure api schema is not stale

kirkbyo:
Looks like [I was wrong](https://github.com/andymatuschak/orbit/pull/208#discussion_r637335418). Adding an annotation like you mentioned does work. I guess when I tested that solution locally I was running a stale build 🙃.  Anyways, I reverted my type annotations changes.

I also added a Github action to make sure that the API schema is always up to date with the latest code changes.

### DIFF
--- a/.github/workflows/ts.yml
+++ b/.github/workflows/ts.yml
@@ -55,6 +55,8 @@ jobs:
           node-version: 14.x
       - name: "install dependencies"
         run: yarn install
+      - name: "run build"
+        run: "yarn build"
       - name: "run generateSchema"
         working-directory: packages/api
         run: "yarn generateSchema"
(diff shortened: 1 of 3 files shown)


.github/workflows/ts.yml
```diff
@@ -44,3 +44,21 @@ jobs:
         run: "yarn build"
       - name: "run test"
         run: "yarn test"
+  stale-api-schema:
+    name: API schema is up to date
```
kirkbyo, line 48: Not crazy about this name. Let me know if you have any preference for other names! This job could also be rolled into the build step.

andymatuschak, line 48: Tricky. Perhaps `Generated API schema matches source interfaces` would convey the expectation slightly more explicitly.

andymatuschak (approved):
Glad to hear that `typescript-json-schema` cooperated after all!

PR #211, opened 2021-05-25
Fix spelling error in homepage docs

jessmartin:
Also ran the rest of the doc through spell-check. That appears to be the only spelling error.


andymatuschak:
How embarrassing. Thank you, Jess!

PR #212, opened 2021-05-25
Add validation to the /taskData endpoint

kirkbyo:
Relates to https://github.com/andymatuschak/orbit/issues/184

Move the `/taskData` to be validated. I figured this work deserved its own PR since there were a couple weird quirks. Hopefully the next endpoints will be a lot smoother and will be done all in one PR.

**How was this tested?**:
I setup a collection of HTTP requests using a [Postman](https://www.postman.com/) collection. Works great! I also went ahead and migrated the test cases I used for the `/actionLogs` endpoint to make sure there are no regressions as I implement changes. It would be nice to eventually integrate postman into CI or have our own HTTP runner.

<details>
<summary>Test cases</summary>

```json
{
   "info":{
      "_postman_id":"1f3abda8-add9-49b9-a79b-47570da6c37c",
      "name":"/taskData",
      "schema":"https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
   },
   "item":[
      {
         "name":"it allows array of ids",
         "event":[
            {
               "listen":"test",
               "script":{
                  "exec":[
                     "pm.test(\"it succeeds\", function() {",
                     "    pm.response.to.not.have.status(400);",
                     "})"
                  ],
                  "type":"text/javascript"
               }
            }
         ],
         "request":{
            "method":"GET",
            "header":[
               
            ],
            "url":{
               "raw":"{{HOST}}/api/taskData?ids=someIdAp&ids=someIdB",
               "host":[
                  "{{HOST}}"
               ],
               "path":[
                  "api",
                  "taskData"
               ],
               "query":[
                  {
                     "key":"ids",
                     "value":"someIdA"
                  },
                  {
                     "key":"ids",
                     "value":"someIdB"
                  }
               ]
            }
         },
         "response":[
            
         ]
      },
      {
         "name":"it allows single id",
         "event":[
            {
               "listen":"test",
               "script":{
                  "exec":[
                     "pm.test(\"it succeeds\", function() {",
                     "    pm.response.to.not.have.status(400);",
                     "})"
                  ],
                  "type":"text/javascript"
               }
            }
         ],
         "request":{
            "method":"GET",
            "header":[
               
            ],
            "url":{
               "raw":"{{HOST}}/api/taskData?ids=someIdA",
               "host":[
                  "{{HOST}}"
               ],
               "path":[
                  "api",
                  "taskData"
               ],
               "query":[
                  {
                     "key":"ids",
                     "value":"someIdA"
                  }
               ]
            }
         },
         "response":[
            
         ]
      },
      {
         "name":"it fails when body is not provided",
         "event":[
            {
               "listen":"test",
               "script":{
                  "exec":[
                     "pm.test(\"it fails with body should be array\", function() {",
                     "    pm.response.to.have.status(400);",
                     "    const expected = [",
                     "        { \"message\": \"body must be array\" }",
                     "    ];",
                     "    const json = pm.response.json();",
                     "    pm.expect(json.errors).to.deep.eq(expected);",
                     "})"
                  ],
                  "type":"text/javascript"
               }
            }
         ],
         "request":{
            "method":"PATCH",
            "header":[
               
            ],
            "url":{
               "raw":"{{HOST}}/api/taskData",
               "host":[
                  "{{HOST}}"
               ],
               "path":[
                  "api",
                  "taskData"
               ]
            }
         },
         "response":[
            
         ]
      },
      {
         "name":"it fails when prompt is not provided",
         "event":[
            {
               "listen":"test",
               "script":{
                  "exec":[
                     "pm.test(\"it fails with required property data\", function() {",
                     "    pm.response.to.have.status(400);",
                     "    const expected = [",
                     "        { \"message\": \"body/0 must have required property 'data'\" }",
                     "    ];",
                     "    const json = pm.response.json();",
                     "    pm.expect(json.errors).to.deep.eq(expected);",
                     "})"
                  ],
                  "type":"text/javascript"
               }
            }
         ],
         "request":{
            "method":"PATCH",
            "header":[
               
            ],
            "body":{
               "mode":"raw",
               "raw":"[\n    {\n        \"id\": \"a\"\n    }\n]",
               "options":{
                  "raw":{
                     "language":"json"
                  }
               }
            },
            "url":{
               "raw":"{{HOST}}/api/taskData",
               "host":[
                  "{{HOST}}"
               ],
               "path":[
                  "api",
                  "taskData"
               ]
            }
         },
         "response":[
            
         ]
      },
      {
         "name":"it fails when id is not provided",
         "event":[
            {
               "listen":"test",
               "script":{
                  "exec":[
                     "pm.test(\"it fails with property id required\", function() {",
                     "    pm.response.to.have.status(400);",
                     "    const expected = [",
                     "        { \"message\": \"body/0 must have required property 'id'\" }",
                     "    ];",
                     "    const json = pm.response.json();",
                     "    pm.expect(json.errors).to.deep.eq(expected);",
                     "})"
                  ],
                  "type":"text/javascript"
               }
            }
         ],
         "request":{
            "method":"PATCH",
            "header":[
               
            ],
            "body":{
               "mode":"raw",
               "raw":"[\n    { \n        \"data\": {\n            \"promptType\": \"clozePrompt\",\n            \"body\": { \"contents\": \"test\", \"attachments\": [] }\n        }\n    }\n]",
               "options":{
                  "raw":{
                     "language":"json"
                  }
               }
            },
            "url":{
               "raw":"{{HOST}}/api/taskData",
               "host":[
                  "{{HOST}}"
               ],
               "path":[
                  "api",
                  "taskData"
               ]
            }
         },
         "response":[
            
         ]
      },
      {
         "name":"it fails when prompt type is invalid",
         "event":[
            {
               "listen":"test",
               "script":{
                  "exec":[
                     "pm.test(\"it fails with no additionalProperties\", function() {",
                     "    pm.response.to.have.status(400);",
                     "    const expected = [",
                     "        { \"message\": \"body/0/data must have required property 'answer'\" },",
                     "        { \"message\": \"body/0/data must have required property 'variants'\" },",
                     "        { \"message\": \"body/0/data/promptType must be equal to one of the allowed values\" },",
                     "        { \"message\": \"body/0/data must match a schema in anyOf\" }",
                     "    ];",
                     "    const json = pm.response.json();",
                     "    pm.expect(json.errors).to.deep.eq(expected);",
                     "})"
                  ],
                  "type":"text/javascript"
               }
            }
         ],
         "request":{
            "method":"PATCH",
            "header":[
               
            ],
            "body":{
               "mode":"raw",
               "raw":"[\n    {\n        \"id\": \"someId\",\n        \"data\": {\n            \"promptType\": \"unknownPromptType\",\n            \"body\": { \"contents\": \"hello world\", \"attachments\": [] }\n        }\n    }\n]",
               "options":{
                  "raw":{
                     "language":"json"
                  }
               }
            },
            "url":{
               "raw":"{{HOST}}/api/taskData",
               "host":[
                  "{{HOST}}"
               ],
               "path":[
                  "api",
                  "taskData"
               ]
            }
         },
         "response":[
            
         ]
      },
      {
         "name":"it succeeds when body is empty",
         "event":[
            {
               "listen":"test",
               "script":{
                  "exec":[
                     "pm.test(\"it succeeds\", function() {",
                     "    pm.response.to.have.status(401);",
                     "})"
                  ],
                  "type":"text/javascript"
               }
            }
         ],
         "request":{
            "method":"PATCH",
            "header":[
               
            ],
            "body":{
               "mode":"raw",
               "raw":"[]",
               "options":{
                  "raw":{
                     "language":"json"
                  }
               }
            },
            "url":{
               "raw":"{{HOST}}/api/taskData",
               "host":[
                  "{{HOST}}"
               ],
               "path":[
                  "api",
                  "taskData"
               ]
            }
         },
         "response":[
            
         ]
      },
      {
         "name":"it succeeds when body has an item",
         "event":[
            {
               "listen":"test",
               "script":{
                  "exec":[
                     "pm.test(\"it fails with no additionalProperties\", function() {",
                     "    pm.response.to.have.status(401);",
                     "})"
                  ],
                  "type":"text/javascript"
               }
            }
         ],
         "request":{
            "method":"PATCH",
            "header":[
               
            ],
            "body":{
               "mode":"raw",
               "raw":"[\n    {\n        \"id\": \"hello\",\n        \"data\": {\n            \"promptType\": \"clozePrompt\",\n            \"body\": { \"contents\": \"hello\", \"attachments\": [] }\n        }\n    }\n]",
               "options":{
                  "raw":{
                     "language":"json"
                  }
               }
            },
            "url":{
               "raw":"{{HOST}}/api/taskData",
               "host":[
                  "{{HOST}}"
               ],
               "path":[
                  "api",
                  "taskData"
               ]
            }
         },
         "response":[
            
         ]
      }
   ]
}
```
</details>

### DIFF
--- a/packages/api/src/validation/ajvAPIValidator.ts
+++ b/packages/api/src/validation/ajvAPIValidator.ts
@@ -23,7 +23,11 @@ export class AjvAPIValidator<T extends AjvSchema> implements APIValidator {
     const ajv = new Ajv({
       allowUnionTypes: true,
       useDefaults: config.mutateWithDefaultValues,
-      // coerce numbers, boolean values AND array
+      // Coerce incoming values to the schema-specified types. When the schema
+      // specifies an array type, it will map x to [x] as needed.
+      //
+      // ex: let ids be defined as an array in the schema
+      // a value of { ids: "idA" } will be coerced to { ids: ["idA"] }
       coerceTypes: "array",
     });
     this.validator = ajv.compile<T>(schema);
--- a/packages/api/src/validation/ajvAPIValidator.ts
+++ b/packages/api/src/validation/ajvAPIValidator.ts
@@ -79,7 +79,10 @@ export class AjvAPIValidator<T extends AjvSchema> implements APIValidator {
   private _createAPIValidationError(error: AjvErrorObject) {
     // Assume the format is /#/properties/[HTTP_PATH]/properties/[VERB]/properties/query/*;
     // strip everything up to query
-    const isolatedInstancePath = error.schemaPath.split("/").slice(6).join("/");
+    const isolatedInstancePath = error.instancePath
+      .split("/")
+      .slice(3)
+      .join("/");
 
     return {
       message: `${isolatedInstancePath} ${error.message}`,
--- a/packages/api/src/validation/ajvAPIValidator.ts
+++ b/packages/api/src/validation/ajvAPIValidator.ts
@@ -23,7 +23,8 @@ export class AjvAPIValidator<T extends AjvSchema> implements APIValidator {
     const ajv = new Ajv({
       allowUnionTypes: true,
       useDefaults: config.mutateWithDefaultValues,
-      coerceTypes: true,
+      // coerce numbers, boolean values AND array
+      coerceTypes: "array",
     });
     this.validator = ajv.compile<T>(schema);
   }
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -23,6 +23,7 @@
     "@withorbit/api": "0.0.1",
     "@withorbit/core": "0.0.1",
     "@withorbit/firebase-support": "0.0.1",
+    "ajv": "^8.5.0",
     "busboy": "^0.3.1",
     "cookie-parser": "^1.4.5",
     "cors": "^2.8.5",
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -82,5 +82,8 @@
       "firebase-functions",
       "firebase-admin"
     ]
+  },
+  "resolutions": {
+    "api/ajv": "^8.5.0"
   }
 }
--- a/packages/backend/src/api/taskData.ts
+++ b/packages/backend/src/api/taskData.ts
@@ -9,12 +9,7 @@ export const listTaskData: TypedRouteHandler<
   "/taskData",
   "GET"
 > = async (request) => {
-  // TODO HACK: Replace this with real query deserialization.
-  const ids = Array.isArray(request.query.ids)
-    ? request.query.ids
-    : [request.query.ids];
-
-  const taskData = await backend.prompts.getPrompts(ids);
+  const taskData = await backend.prompts.getPrompts(request.query.ids);
 
   return {
     json: {
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -11,7 +11,7 @@
     "generateSchema": "typescript-json-schema src/orbitAPI.ts ValidatableSpec -o src/orbitAPISchema.json --noExtraProps --required --ignoreErrors --strictNullChecks"
   },
   "dependencies": {
-    "ajv": "8.4.0"
+    "ajv": "^8.5.0"
   },
   "devDependencies": {
     "@types/jest": "^26.0.23",
--- a/packages/api/src/orbitAPI.ts
+++ b/packages/api/src/orbitAPI.ts
@@ -31,18 +31,6 @@ type SpecLegacy = {
     };
   };
 
-  "/taskData": {
-    GET: {
-      query: { ids: PromptID[] };
-      response: ResponseList<"taskData", PromptID, Prompt>;
-    };
-
-    PATCH: {
-      body: { id: PromptID; data: Prompt }[];
-      response: void;
-    };
-  };
-
   "/attachments": {
     POST: {
       contentType: "multipart/form-data";
@@ -95,6 +83,17 @@ export type ValidatableSpec = {
         id: ActionLogID;
         data: ActionLog;
       }[];
+      response?: null;
+    };
+  };
+
+  "/taskData"?: {
+    GET?: {
+      query: { ids: PromptID[] };
+      response?: ResponseList<"taskData", PromptID, Prompt>;
+    };
+    PATCH?: {
+      body: { id: PromptID; data: Prompt }[];
       response?: null;
     };
   };
--- a/packages/api/src/orbitAPISchema.json
+++ b/packages/api/src/orbitAPISchema.json
@@ -2,6 +2,70 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "additionalProperties": false,
     "definitions": {
+        "ApplicationPrompt": {
+            "additionalProperties": false,
+            "properties": {
+                "promptType": {
+                    "enum": [
+                        "applicationPrompt"
+                    ],
+                    "type": "string"
+                },
+                "variants": {
+                    "items": {
+                        "$ref": "#/definitions/QAPromptContents"
+                    },
+                    "type": "array"
+                }
+            },
+            "required": [
+                "promptType",
+                "variants"
+            ],
+            "type": "object"
+        },
+        "AttachmentIDReference": {
+            "additionalProperties": false,
+            "properties": {
+                "byteLength": {
+                    "type": "number"
+                },
+                "id": {
+                    "type": "string"
+                },
+                "type": {
+                    "enum": [
+                        "image"
+                    ],
+                    "type": "string"
+                }
+            },
+            "required": [
+                "byteLength",
+                "id",
+                "type"
+            ],
+            "type": "object"
+        },
+        "ClozePrompt": {
+            "additionalProperties": false,
+            "properties": {
+                "body": {
+                    "$ref": "#/definitions/PromptField"
+                },
+                "promptType": {
+                    "enum": [
+                        "clozePrompt"
+                    ],
+                    "type": "string"
+                }
+            },
+            "required": [
+                "body",
+                "promptType"
+            ],
+            "type": "object"
+        },
         "IngestActionLog": {
             "additionalProperties": false,
             "properties": {
@@ -55,6 +119,25 @@
             },
             "type": "object"
         },
+        "PromptField": {
+            "additionalProperties": false,
+            "properties": {
+                "attachments": {
+                    "items": {
+                        "$ref": "#/definitions/AttachmentIDReference"
+                    },
+                    "type": "array"
+                },
+                "contents": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "attachments",
+                "contents"
+            ],
+            "type": "object"
+        },
         "PromptProvenanceType": {
             "enum": [
                 "anki",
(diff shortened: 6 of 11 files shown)


packages/backend/package.json
```diff
@@ -23,6 +23,7 @@
     "@withorbit/api": "0.0.1",
     "@withorbit/core": "0.0.1",
     "@withorbit/firebase-support": "0.0.1",
+    "ajv": "^8.5.0",
```
kirkbyo, line 26: **tldr**: `ajv` is the wrong version when being used on the backend. Results in weird errors. Listing `ajv` as a dependency of the backend resolves the issue.

The situation: it looks like a dependency of another package is using `ajv` but version `6.12.6`. This becomes problematic since the `api` package is using version `8.5.0`. Thus, the file system is setup like:

```
- node_modules
  - ajv (6.12.6)
- packages
  - api
    - node_modules
      - ajv (8.5.0)
  ```

My understanding was that the backend package would use `ajv` version 8.5.0 however this is not the case. The backend is using version of 6.12.6 when being ran in the emulator.

I only noticed after reading the [migration guide for ajv](https://ajv.js.org/v6-to-v8-migration.html#potential-migration-difficulties) which marks `dataPath` as deprecated, yet this property is still appearing in our error objects when used on the backend. I only really noticed this issue after spending quite some time debugging some weird edge case that looks to be resolved in later versions of the package. 

It is worth noting that the `ajvAPIValidator.test.ts` is using the correct version of the package.

Listing `ajv` directly in the backend works, but this can't be the best solution.  I don't have too much experience with yarn monorepo's and versioning, so I have had some trouble debugging this issue further.

File: packages/api/src/validation/ajvAPIValidator.ts [hunk 2]
```diff
@@ -23,7 +23,8 @@ export class AjvAPIValidator<T extends AjvSchema> implements APIValidator {
     const ajv = new Ajv({
       allowUnionTypes: true,
       useDefaults: config.mutateWithDefaultValues,
-      coerceTypes: true,
+      // coerce numbers, boolean values AND array
```
andymatuschak, line 26: It's great that this feature exists! Despite the comment, I still had to look up the documentation to understand what this does. Maybe: "Coerce incoming values to the schema-specified types. When the schema specifies an array type, will map `x` to `[x]` as needed."

packages/backend/package.json
andymatuschak, line 26: Ah, yes. I've run into a few of these issues before. This being a monorepo makes it more confusing, but the underlying situation is not actually all that different from the troubles of a normal Node project in which you depend on `a` and `b@2.0`, while `a` internally depends on `b@1.0`. The Node resolver expects to resolve the same package for `require` calls in your module as well as in `a`: you've got to pick just one to load.

In such situations, you must either use the old version of `b` yourself until `a` upgrades, or force `a` to use `b@2.0` via [selective resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/). The latter may or may not work, depending on the nature of the breaking change. In practice, upgrading `a` sometimes suffices, since they might have updated their dependency on `b`.

This solution amounts to the latter: by installing the newer version of `ajv` in `backend/node_modules`, Node will preferentially resolve that one over the one at the monorepo root. This works because it happens that aside from `eslint`, none of `backend`'s dependencies depend on an older version of `ajv`. And it happens to work when running tests in `api` because none of the dependencies of that module depend on an older version of `ajv` either. We're really just getting lucky there, I guess!

Can we do better? Well... in our project, we have _many_ libraries which depend on `ajv` 6.x:
```
[ui (master)]$ yarn why ajv
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "ajv"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "ajv@6.12.6"
info Has been hoisted to "ajv"
info Reasons this module exists
   - "workspace-aggregator-e8a52493-a98e-40c9-a67b-5c8d3ba1a303" depends on it
   - Hoisted from "_project_#eslint#ajv"
   - Hoisted from "_project_#webpack#ajv"
   - Hoisted from "_project_#schema-utils#ajv"
   - Hoisted from "_project_#terser-webpack-plugin#schema-utils#ajv"
   - Hoisted from "_project_#webpack#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#backend#eslint#ajv"
   - Hoisted from "_project_#@withorbit#ui#eslint#ajv"
   - Hoisted from "_project_#eslint#table#ajv"
   - Hoisted from "_project_#eslint#@eslint#eslintrc#ajv"
   - Hoisted from "_project_#@withorbit#ui#react-docgen-typescript-loader#@webpack-contrib#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#app#@expo#webpack-config#webpack#ajv"
   - Hoisted from "_project_#@withorbit#app#webpack#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#web-component#webpack#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#app#@expo#webpack-config#worker-loader#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#app#@expo#webpack-config#mini-css-extract-plugin#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#app#@expo#webpack-config#webpack#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#app#@expo#webpack-config#babel-loader#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#app#@expo#webpack-config#terser-webpack-plugin#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#app#@expo#webpack-config#url-loader#schema-utils#ajv"
   - Hoisted from "_project_#@withorbit#ui#@storybook#react#@storybook#core#postcss-loader#schema-utils#ajv"
info Disk size without dependencies: "1.12MB"
info Disk size with unique dependencies: "1.86MB"
info Disk size with transitive dependencies: "1.9MB"
info Number of shared dependencies: 5
=> Found "@withorbit/api#ajv@8.4.0"
info This module exists because "_project_#@withorbit#api" depends on it.
=> Found "exegesis#ajv@6.12.5"
info This module exists because "_project_#@withorbit#app#firebase-tools#exegesis-express#exegesis" depends on it.
info Disk size without dependencies: "1.12MB"
info Disk size with unique dependencies: "1.86MB"
info Disk size with transitive dependencies: "1.9MB"
info Number of shared dependencies: 5
=> Found "har-validator#ajv@6.12.5"
info This module exists because "_project_#@withorbit#firebase-support#@firebase#rules-unit-testing#request#har-validator" depends on it.
info Disk size without dependencies: "1.12MB"
info Disk size with unique dependencies: "1.86MB"
info Disk size with transitive dependencies: "1.9MB"
info Number of shared dependencies: 5
```

And I see that `eslint@master` [still references](https://github.com/eslint/eslint/blob/master/package.json) `ajv` 6.x, so updating our dependencies won't fix the issue. I guess we're stuck! There's an issue about updating in eslint [here](https://github.com/eslint/eslint/issues/13888). (Boy, some of the comments from the `ajv` people about the churn they're dealing with in the JSON Schema spec scare me!)

The only potential improvement I see would be to declare our intent somewhat more explicitly here, so that we make clear that `backend` doesn't directly depend on `ajv`: it just needs the version resolved a certain way. You should be able to do that by declaring a selective resolution instead of an explicit dependency, i.e.:

```
"resolutions": {
  "api/ajv": "^8.5.0"
}
```

packages/backend/src/api/taskData.ts
```diff
@@ -9,12 +9,7 @@ export const listTaskData: TypedRouteHandler<
   "/taskData",
   "GET"
 > = async (request) => {
-  // TODO HACK: Replace this with real query deserialization.
-  const ids = Array.isArray(request.query.ids)
-    ? request.query.ids
-    : [request.query.ids];
-
-  const taskData = await backend.prompts.getPrompts(ids);
+  const taskData = await backend.prompts.getPrompts(request.query.ids);
```
andymatuschak, line 12: Rad!!!

andymatuschak (approved):

packages/api/src/validation/ajvAPIValidator.ts
kirkbyo: I like that. I will update!

packages/backend/package.json
kirkbyo: Oh wow, TIL. Thank you for the detailed explanation. That makes a lot more sense now. I will go ahead and add the resolutions to the `package.json`.

> (some of the comments from the ajv people about the churn they're dealing with in the JSON Schema spec scare me!)

Yeah same... I hope by keeping our spec relatively "vanilla" (i.e. no custom keywords or plugins) spec changes won't be breaking 🤞🏻. For better or worse `typescript-json-schema` keeps us rather constricted from getting ahead of ourselves.

PR #213, opened 2021-05-26
Convert /taskStates, /attachments and /attachments/:id routes to be validated

kirkbyo:
Relates to #184

Converted `/taskStates`, `/attachments` and `/attachments/:id` to all be backed by the validation system.

**A few notes on the changes**:
- `/attachments` includes a `contentType` field. Rather then just making this an optional field, I figured the other routes could benefit from having this validation as well. Thus, I extended their schema to include a `contentType: JSONContentType` field.
- To support validation for `/attachments/:id`, the validation mechanism was moved from a middleware to be done within the typed router setup. This was the result of the route params not being available in the middleware (https://github.com/expressjs/express/issues/3177). 
   - This also has the dual effect of extending the validation for the `/attachments` route since the body has seen serialized by the time it reaches the validation code in the typed router, thus the file size and type can be validated.

**How this was tested**
I extended my postman collections to include test cases for the new routes and ran the old test cases for the other routes to ensure there were no regressions.

You can see all the test case cases here: https://www.postman.com/orbit-postman/workspace/orbit/overview


packages/api/src/orbitAPI.ts
```diff
@@ -12,58 +12,32 @@ import { BlobLike } from "./genericHTTPAPI";
...
-  "/taskStates": {
-    GET: {
+export type ValidatableSpec = {
+  "/taskStates"?: {
+    GET?: {
       query:
         | ({
+            /**
+             * @minimum 1
+             * @TJS-type integer
```
kirkbyo, line 22: Cannot specify a default here due to a [limitation with `ajv`](https://ajv.js.org/guide/modifying-data.html#assigning-defaults). Defaults cannot be assigned within a `anyOf`.

packages/api/src/orbitAPI.ts
```diff
@@ -93,13 +68,71 @@ export type ValidatableSpec = {
...
+
+  "/attachments"?: {
+    /**
+     * encode with multipart/form-data, with the file in part named "file"
+     * make sure to include Content-Type heading for your attachment
+     * returns application/json encoded ResponseObject<"attachmentIDReference", AttachmentID, AttachmentIDReference>
+     */
+    POST?: {
+      // NOTE: Content-type must use regex to be validated since additional data is usually
+      // included alongside the contentType (i.e. the form-data length)
```
andymatuschak, line 85: Ah, juicy! Makes sense. It took me a moment to understand this comment: perhaps "…since additional data, like the form-data length, is usually appended to the MIME type" would be clearer?

packages/api/src/orbitAPI.ts
```diff
@@ -93,13 +68,71 @@ export type ValidatableSpec = {
...
+    };
+  };
 };
 
-export type Spec = RequiredSpec<ValidatableSpec> & SpecLegacy;
+/**
+ * @additionalProperties true
+ */
+export interface FileUploadBlob extends BlobLike {
+  type: "image/png" | "image/jpeg" | "image/svg+xml";
```
andymatuschak, line 119: Could this be set to `AttachmentMimeType` to avoid the duplication?

packages/api/src/validation/ajvAPIValidator.ts
```diff
@@ -45,13 +45,22 @@ export class AjvAPIValidator<T extends AjvSchema> implements APIValidator {
   }
 
   private validate(
-    { method, path, query, body }: APIValidatorRequest,
+    { method, path, contentType, query, body, params }: APIValidatorRequest,
     response: unknown,
   ) {
+    // helper to make sure that objects are not empty
+    const isDefined = (obj: Record<string, unknown> | undefined) => {
```
andymatuschak, line 52: Nit: the name here implies a boolean return value. This is really mapping empty/falsy objects to `undefined`. Might I suggest a name like `undefinedIfEmpty` instead?

Extremely nitpicky nit, which you should feel totally free to ignore: I think it's clearer to use `null` in situations like this than `undefined`, because the two have subtly different meanings. `null` represents the intentional and definitive absence of a value, whereas `undefined` values are automatically assigned by the environment when an expression extends outside its domain of definition, like "a variable which has been declared but not initialized," "a valid object projection on an uninitialized key", "the return value of a void function", etc. Here we're defining a function with an intentional semantic of "emptiness" (i.e. objects with no keys map to emptiness for our purposes here).

packages/api/src/validation/apiValidator.ts
```diff
@@ -1,8 +1,10 @@
 export type APIValidatorRequest = {
   method: string;
   path: string;
-  query?: unknown;
-  body?: unknown;
+  contentType?: string;
+  params?: { [key: string]: unknown };
+  query?: { [key: string]: unknown };
+  body?: any;
```
andymatuschak, line 7: Just curious: can't keep `unknown` here?

packages/backend/src/api/taskStates.ts
```diff
@@ -17,21 +17,9 @@ export const listTaskStates: TypedRouteHandler<
       if ("ids" in query) {
         promptStates = await backend.promptStates.getPromptStates(
           userID,
-          Array.isArray(query.ids) ? query.ids : [query.ids],
+          query.ids,
         );
       } else {
-        // TODO HACK: need real type-safe parsing...
```
andymatuschak, line 23: Great to see this stuff removed.

packages/core/src/promptState/promptState.ts
```diff
@@ -12,6 +12,11 @@ export interface PromptState<PT extends PromptTask = PromptTask> {
   taskMetadata: PromptTaskMetadata;
 
   lastReviewTimestampMillis: number;
+  /**
+   * @TJS-type object
+   * @additionalProperties true
+   * @nullable
```
andymatuschak, line 18: Curious: why does nullability need to be declared when the type is `| null`? Just a quirk of `typescript-json-schema`?

andymatuschak (approved):
Looking great!

packages/api/src/validation/ajvAPIValidator.ts
kirkbyo: Good point, makes sense. I'll go ahead and use null instead!

packages/api/src/validation/apiValidator.ts
kirkbyo: Oops, yes we can. I'll revert this change.

packages/core/src/promptState/promptState.ts
kirkbyo: Yep exactly, a small quirk of `typescript-json-schema`. When we specify the `@TJS-type` it'll completely override any type definition, leaving the `| null` to be forgotten in the process.

packages/api/src/orbitAPI.ts
andymatuschak: Occurs to me: I suppose the pattern should actually be `^multipart/form-data` (i.e. enforcing the string as prefix), right?

andymatuschak:
(Should have mentioned: I'll leave it to you to merge when ready!)

Issue #195, opened 2021-04-26
Remove firebase/auth from clients, use our own APIs instead


andymatuschak:
I'd like to compartmentalize Firebase as much as possible. Right now, the only significant remaining "leakage" of our usage to clients is `firebase/auth`, which we use to sign in, persist credentials, etc. We should instead make requests against our own backend APIs, which can call the appropriate Firebase APIs. This will save bundle size (that library is huge!), but it'll also make it easier when we eventually make self-hosting variants, etc.

Bonus points for migrating existing users' stored `firebase/auth` credentials to whatever new persistence scheme we implement.

seanwessmith:
Do you still view Firebase behind API as best potential option? Potential new API endpoints:

  POST /auth/register       → backend calls Firebase Admin
  POST /auth/login          → backend calls Firebase Admin
  POST /auth/check-email    → backend calls Firebase Admin
  POST /auth/reset-password → backend calls Firebase Admin

  Pros: Minimal risk, reuses existing infrastructure, ~100KB bundle savings
  Cons: Still coupled to Firebase on backend
  ---
Or could go opensource with SuperTokens.

  Replace Firebase auth entirely with https://supertokens.com, an open-source auth solution that:

  - Node SDK
  - Migration tooling for existing Firebase users
  - ~15KB gzipped

  Pros: Open source, smaller bundle
  Cons: New infrastructure component, migration effort

  ---

  Happy to help with implementation if there's interest in either approach.

PR #214, opened 2021-05-27
Web extension

growthepie:
# Motivation

Whenever I come across something interesting on the web I have the urge to incorporate that concept into my orbit. This browser extension would allow me to do so by making it possible to create Orbit prompts on any website.

Screen Shot 2021-05-28 at 00 01 17

For now, I've worked on the code structure and built some basic UI.

## Questions

- How can I create new prompts using the `@withorbit/api-client` package? Couldn't find a clear way to do this.
- Is there a standardized way to extract the **Page title** from the current site?
- How should authentication work? For now, I'd be fine with setting an `ORBIT_TOKEN` and saving it to local storage. (Might be worth waiting for #195?)
- What should be the license?

## Caveats

- Only Chrome for now
- Attachments not supported
- Missing larger icons (*48, 128*)
- Activation using keyboard shortcut doesn't work (`_execute_action`)

*Not sure if you'd like this to be part of the repo — just putting this out there.*


andymatuschak:
Very cool! Some replies to your questions:

> How can I create new prompts using the @withorbit/api-client package? Couldn't find a clear way to do this.

I'm sorry that the APIs aren't documented yet. As it happens, they're still in flux and not quite ready for adoption, so—beware!! But if you'd like to keep playing in the meantime, you'd create a new prompt by `PATCH`ing `/taskData` with the prompt contents and ID (see orbitAPI.ts).

> Is there a standardized way to extract the Page title from the current site?

Not that I know. You can see how the web component does this at /packages/api/web-component/src/metadataMonitor.ts.

> How should authentication work? For now, I'd be fine with setting an ORBIT_TOKEN and saving it to local storage. (Might be worth waiting for #195?)

This looks like a great candidate for an OAuth client! i.e. so that an Orbit user can indicate that this plugin should be able to read/write their account. Filed: https://github.com/andymatuschak/orbit/issues/216

> What should be the license?
> Not sure if you'd like this to be part of the repo — just putting this out there.

I'm excited that you're experimenting with this! Just for expediency, I'd suggest that you continue iterating on this package in a separate repo. (I'd be happy to keep providing comments / help there) If it evolves into something production-ready, we can figure out what it might mean to make it "official".

PR #215, opened 2021-06-01
Add warning for unsupported old browsers

### LINKED ISSUE #166: Add warning for old browsers
Browsers which aren't supported by our browserslist should see a warning in the UI.

kirkbyo:
Closes #166 

**How**
Using the default `browserslist` config a regex string is generated to verify that the userAgent of the browser at runtime is considered "supported". In the event that the browser is not supported the following banner is shown.

Screen Shot 2021-06-01 at 10 20 17 AM

**How this was tested**
I mocked the browsers `userAgent` by adding the following script at the top of the body of `packages/app/web/index.html`:

```html
<script>
navigator.__defineGetter__('userAgent', function(){
  return 'Mozilla/5.0 (Linux; Android 5.1.1; SM-G928X Build/LMY47X) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.83 Mobile Safari/537.36'
});
</script>
```

**Notes on the approach**
- Instead of using `browserslist-useragent-regexp` we could use `browserslist-useragent`. This would require importing a package which might be tricky without having the banner within the React bundle.
- I opted to include the banner within the root `index.html`  instead of within React in the case that React is not supported.


packages/app/web/index.html
```diff
@@ -45,6 +45,7 @@
 
       body {
         display: flex;
+        flex-direction: column;
```
kirkbyo, line 48: Any preference on having the banner block level vs. `position: absolute` instead?

andymatuschak, line 48: I think this is the right thing to do! The alternative would often produce strange overlaps.

andymatuschak (approved):
Great work, thanks; please merge when ready!

Issue #179, opened 2021-01-28
Improve embedded login UX when credential already present


andymatuschak:
Right now, if you sign into an embedded prompt when you're already signed in, the popup window will be able to read the credential, and then it'll immediately close, which is jarring and surprising. We should improve this UX (maybe a "Continue with x@y.com?")

kirkbyo:
I am looking to take this on but I am having trouble reproducing the issue. Here are the steps I have tried:

1. Fresh DB (no users), spin up the web-components test page (`localhost:5000`) and complete the first set of prompts.
2. After completing the prompts, create a new user.
3. Complete the second test. This is where I would have expected the "flash" of the auth window to appear, but I did not see it.
4. Open a new browser and complete the third test. Sign into the user created during step 2. 

[Demo of me walking through the steps above](https://www.loom.com/share/4b71a6dde5f245f4badedfdf4d36d615)

Let me know if I am missing something or miss-understanding the issue.

andymatuschak:
I really should have included reproduction steps; I'm sorry! The problem occurs when an embedded review area (which is a third-party iframe) is prevented from accessing the stored credentials, while a popup (a first-party window) is permitted. When this occurs depends on your browser and security settings, but in general it'll happen if you _move between_ sites which use Orbit, and in some circumstances between sessions on the same site.

Unfortunately, this can be tricky to see locally because `localhost:5000` and `localhost:19006` are considered the same domain (at least by Chrome), and so the iframe isn't sandboxed like a third party.

1. Set Chrome to block third-party cookies.
2. Fresh DB (no users), spin up the web-components test page but visit http://127.0.0.1:5000 instead of `localhost:5000` and complete the first set of prompts.
3. After completing the prompts, create a new user. At this point, other review areas on the same page should use that credential.
4. Refresh the page. Your progress should not be loaded.
5. Complete the first review area again. Enter `test@test.com` into the email field.
6. You should see the auth popup appear and immediately disappear.

Safari exhibits the same behavior with its _default_ settings (i.e. without third-party cookie blocking set) if you navigate to two different embedding hosts. Using 127.0.0.1 and your network IP (e.g. 192.168.0.167) might be enough to exhibit the issue.

PR #217, opened 2021-06-01
Add 'continue with user' option when credentials are already present

kirkbyo:
Closes https://github.com/andymatuschak/orbit/issues/179

<p align="center">

</p>

**What**
When the credentials are already present on the `iframe`'s domain, a "Continue with <email>" option is shown to confirm with the user that they are already authenticated.

<details>
<summary>Demo of Before and After behaviour</summary>

| Before | After |
| --- | --- |
| old_flow | new_flow |

</details>

**How**
If the `userRecord` is received before the user has entered their authentication credentials, we'll "double check" that they want to use the account we found. I added a new "ContinueWithUser` component to handle this behaviour to avoid bloating `SigninForm` further with conditionals. 

**Notes**
I will admit this new flow feels a bit "wonky". It feels weird to ask the user to "continue" without presenting them another option. I attempted to add a button that allows the user to "Use a different account", but to do this flow properly would involve reworking how the `overrideEmailAddress` is used right now. If we did want this behaviour, I'd be happy to take that on in a separate PR!

### DIFF
--- a/packages/app/src/signIn/SignInScreen.tsx
+++ b/packages/app/src/signIn/SignInScreen.tsx
@@ -1,4 +1,9 @@
-import { SignInForm, SignInFormProps, styles } from "@withorbit/ui";
+import {
+  ContinueWithUser,
+  SignInForm,
+  SignInFormProps,
+  styles,
+} from "@withorbit/ui";
 import React from "react";
 import { ActivityIndicator, Alert, Platform, View } from "react-native";
 import { AuthenticationClient } from "../authentication";
@@ -80,6 +85,10 @@ export default function SignInScreen() {
   const [isPendingServerResponse, setPendingServerResponse] = React.useState(
     false,
   );
+  const [
+    shouldDoubleCheckAccountStatus,
+    setShouldDoubleCheckAccountStatus,
+  ] = React.useState<"requiresUserApproval" | "approved" | null>(null);
 
   // If we have an override email address, figure out whether that account exists.
   React.useEffect(() => {
@@ -95,11 +104,22 @@ export default function SignInScreen() {
   }, [authenticationClient, overrideEmailAddress]);
 
   const userRecord = useCurrentUserRecord(authenticationClient);
+
+  React.useEffect(() => {
+    if (userRecord && shouldDoubleCheckAccountStatus === null) {
+      // a user record was received before the user entered their credentials
+      // lets double check they want to use this account
+      setShouldDoubleCheckAccountStatus("requiresUserApproval");
+    }
+  }, [userRecord, shouldDoubleCheckAccountStatus]);
+
   React.useEffect(() => {
     if (userRecord) {
       const tokenTarget = getCurrentLoginTokenTarget();
       if (tokenTarget) {
-        sendTokenToTargetAndClose(authenticationClient, tokenTarget);
+        if (shouldDoubleCheckAccountStatus === "approved") {
+          sendTokenToTargetAndClose(authenticationClient, tokenTarget);
+        }
       } else {
         if (Platform.OS === "web") {
           // TODO: redirect somewhere useful outside the embedded case
@@ -114,11 +134,12 @@ export default function SignInScreen() {
         }
       }
     }
-  }, [authenticationClient, userRecord]);
+  }, [authenticationClient, userRecord, shouldDoubleCheckAccountStatus]);
 
   const onLogin = React.useCallback(
     async (email, password) => {
       setPendingServerResponse(true);
+      setShouldDoubleCheckAccountStatus("approved");
 
       try {
         switch (formMode) {
@@ -155,6 +176,10 @@ export default function SignInScreen() {
     [authenticationClient],
   );
 
+  const onContinueWithUser = React.useCallback(() => {
+    setShouldDoubleCheckAccountStatus("approved");
+  }, []);
+
   return (
     <View
       style={{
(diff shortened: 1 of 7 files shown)


andymatuschak:
Nice work, @kirkbyo! I'll comment at a high level before digging into the code changes.

At first, I thought: oh, boy, I guess we really have to make the sign in flow much more fleshed out to make this reasonable. To your point, it's weird not to have a second option!

But then I realized: hey, wait. In this flow, the user has _already entered_ their email address in the prior screen. This is quite different from other OAuth flows, where they really need to clarify that you _intend_ to sign in using the saved credential. In our case, if the saved credential's email address matches the one you entered, that's what you mean! This lets us simplify the UI language, and also means we don't need an alternative option.

So if the user's signed in, I think the secondary text can say: "Welcome back, `EMAIL`. We signed you in using a saved password." and the button can say "Continue".

Now, the other thing this makes me realize is that if the stored credential _doesn't_ match the email address they entered, we should discard it, sign out, and show the normal login interface. After all, they asked to sign in with a different address! A fancier flow might say "hey, wait, did you want to sign in with this account instead?" (people often forget which email address they use for things), but I think we can skip that for now.

andymatuschak:
Oh, and probably we should `focus()` the button on mount, so that people can just hit return / space / whatever to continue.

andymatuschak (changes requested):
(Oops, should have made my previous comment as a review to make it show up in your inbox appropriately, sorry)

kirkbyo:
Makes sense! I have updated the auth flow with the changes you mentioned:
<p align="center">
Screen Shot 2021-06-02 at 12 11 22 PM
</p>

<details>
<summary>Demo</summary>

### Flow for when the `userRecord` and the `overrideEmailAddress` match
new_flowV2

### Flow for when the `userRecord` and the `overrideEmailAddress` __do not__ match

different_email

</details>

packages/ui/src/components/ContinueWithUser.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { StyleSheet, Text, View } from "react-native";
+import { colors, type, layout } from "../styles";
+import Button from "./Button";
+import { IconName } from "./IconShared";
+import Logo from "./Logo";
+import Spacer from "./Spacer";
+
+export interface ContinueWithUserProps {
+  colorPalette: colors.ColorPalette;
+  email: string | null;
```
andymatuschak, line 11: Would you mind making this field required and throwing in the parent if `email` ends up `null`? If that happens I'd like to make sure it sounds alarms in Sentry. They won't be able to proceed anyway!

packages/app/src/signIn/SignInScreen.tsx
```diff
@@ -95,11 +104,29 @@ export default function SignInScreen() {
   }, [authenticationClient, overrideEmailAddress]);
 
   const userRecord = useCurrentUserRecord(authenticationClient);
+
+  React.useEffect(() => {
```
andymatuschak, line 108: I think this effect can be merged into the the one on line 123, which might make this state machine a bit clearer. Boy, effect hooks do start to fall apart when you get a bunch of them together like this.

packages/app/src/signIn/SignInScreen.tsx
```diff
@@ -95,11 +104,29 @@ export default function SignInScreen() {
...
+      setShouldDoubleCheckAccountStatus("requiresUserApproval");
+    }
+  }, [userRecord, shouldDoubleCheckAccountStatus]);
+
+  React.useEffect(() => {
+    if (userRecord && userRecord.emailAddress !== overrideEmailAddress) {
+      authenticationClient.signOut();
+      setShouldDoubleCheckAccountStatus(null);
+    }
+  }, [authenticationClient, userRecord, overrideEmailAddress]);
```
andymatuschak, line 121: Ditto this one: maybe best to make one big effect which amounts to "we got a new user state (or account approval); let's see what needs doing." Wish I could see a better way to simplify this. I suppose we could reify the state machine more explicitly, but I'm not sure it's worth it.

packages/app/src/signIn/SignInScreen.tsx
```diff
@@ -80,6 +85,10 @@ export default function SignInScreen() {
   const [isPendingServerResponse, setPendingServerResponse] = React.useState(
     false,
   );
+  const [
+    shouldDoubleCheckAccountStatus,
+    setShouldDoubleCheckAccountStatus,
+  ] = React.useState<"requiresUserApproval" | "approved" | null>(null);
```
andymatuschak, line 91: Nit: this name implies a boolean but it's really a more complex state. Perhaps `automaticLoginState`?

andymatuschak (approved):
Thanks for this, Ozzie! This patch reminds me that we’re probably just creating pain for ourselves by separating `ui` into a separate module like this. Perhaps at some point we’ll want to just move it into `app`. Or maybe just simple components like `Button` should be in a separate `ui` module? Or maybe a separate `style` module just for static bits like color / type hierarchy? Not sure.

Anyway, I made a few notes you might consider, but feel free to ignore all. Merge when you like.

kirkbyo:
Done! I will merge once the tests go green.

> This patch reminds me that we’re probably just creating pain for ourselves by separating ui into a separate module like this. Perhaps at some point we’ll want to just move it into app. Or maybe just simple components like Button should be in a separate ui module? Or maybe a separate style module just for static bits like color / type hierarchy? Not sure.

My two cents: I personally like the idea of separating the UI from the app since it forces you to not encapsulate API logic within the UI and allows for testable components like `SigninForm`. However, I do think the folder could be separated into design system components (i.e. `Button`) vs. views (`SigninForm` or `ReviewArea`).  I worry that moving the design system bits into a separate package would become tedious at this stage since we'd now have two storybook instances.

andymatuschak:
Those are all good points! 👍

Issue #165, opened 2020-12-10
Keyboard shortcuts for review sessions


andymatuschak:
We should have keyboard shortcuts for "reveal", "remembered", and "forgotten." 

These should be (at least relatively) resilient to a) changes to the "remembered" / "forgotten" label wording; b) adding a third or fourth action.

It should probably be possible to access all keyboard shortcuts one-handed.

We should probably display some affordance in the UI (only if a physical keyboard is present) to hint the key. No strong feelings at this moment: we could start with a tooltip or a hover element.

growthepie:
I can work on this!

I was thinking <kbd>Space</kbd> for both *reveal* & *remembered* and <kbd>F</kbd> for *forgotten*. Reasons:
- Fast review process: most common action is *reveal* -> *remembered*, which would mean double spaces
- One-handed: only requires left thumb and index finger
- Even if *forgotten* will be changed to something else, <kbd>F</kbd> is still relevant because of the [Press F to Pay Respects](https://knowyourmeme.com/memes/press-f-to-pay-respects) meme

Should this be implemented in the `@withorbit/ui` package or in the `@withorbit/app`?

andymatuschak:
Thank you for offering, @growthepie!

One interesting consideration here: in the future, I'd like Orbit to support microtasks other than memory prompts, which might have different feedback buttons. For instance, a lightweight exercise ("improvise a melody in pentatonic minor") might have similar buttons with different titles ("went well" / "needs more practice"). A gratitude activity ("visualize the surprise party your friends threw you") might have three buttons ("repeat with same interval", "see more often", "see less often").

It'd be neat to choose keyboard shortcuts which could "grow" in this way. It's probably reasonable to assume that the buttons will continue to be arranged in a sequence (e.g. in a row) and that there's (often) a default button.

Given that, how about `space` for "reveal" and "default action" (i.e. "remembered"), and then `1` / `2` for "forgotten" / "remembered" respectively? Still one-handed, which is nice.

andymatuschak:
(Oh, regarding where to implement: I'd suggest `ReviewArea.tsx`!)

growthepie:
Great — I love the idea of microtasks! Sometimes my prompts consist of just a link to a Khan Academy unit test. Maybe those could be considered microtasks as well and could use custom actions.

> It'd be neat to choose keyboard shortcuts which could "grow" in this way.

Totally agree, let's do it this way! Will work on a PR.

PR #218, opened 2021-06-02
Add keyboard shortcuts for review sessions

growthepie:
Added an `useKey` hook and implemented the default actions (<kbd>Space</kbd>: *Reveal* -> *Remembered*) and <kbd>1</kbd>/<kbd>2</kbd> for the list of actions (*Forgotten* and *Remembered* for now).

Currently there is no hint for the shortcuts. I think we could show it while hovering the buttons.

Also, noticed there is logic for `pendingOutcomeChange` which is called when the user is interacting with one of the actions. Should we call `pendingOutcomeChange` on `keydown` and only call the handlers on `keyup` to make use of this API? If yes, we should probably move the shortcut handling logic inside `Button.tsx`.

> Note: Storybook focuses the sidebar when pressing <kbd>1</kbd> (just keep that in mind while testing).

#165

### DIFF
--- a/packages/ui/src/components/ReviewArea.tsx
+++ b/packages/ui/src/components/ReviewArea.tsx
@@ -6,7 +6,6 @@ import { colors, layout } from "../styles";
 import { Size } from "../util/Size";
 import Card, { CardProps } from "./Card";
 import FadeView from "./FadeView";
-import useKey from "./hooks/useKey";
 import useLayout from "./hooks/useLayout";
 import usePrevious from "./hooks/usePrevious";
 import { useTransitioningValue } from "./hooks/useTransitioningValue";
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/ReviewArea.tsx
```diff
@@ -219,6 +220,34 @@ export default React.memo(function ReviewArea({
...
+    },
+    [currentItem, onMark],
+  );
+
+  // default action (Space)
+  useKey(" ", () => {
+    if (!isShowingAnswer) {
+      setShowingAnswer(true);
+    } else {
+      onMemoizedMark(PromptRepetitionOutcome.Remembered);
```
andymatuschak, line 238: Would you mind making this reference `actions[1]` so that the two actions don't accidentally drift apart?

packages/ui/src/components/ReviewArea.tsx
```diff
@@ -219,6 +220,34 @@ export default React.memo(function ReviewArea({
       ? currentItem?.colorPalette ?? items[items.length - 1].colorPalette
       : colors.palettes.red;
 
+  const onMemoizedMark = useCallback(
```
andymatuschak, line 223: Do you think it would be cleaner to move these changes into `ReviewButtonBar` so that all the "button knowledge" is in one place? I know that I suggested `ReviewArea` initially, but now that I look at this, I see that this code makes implicit assumptions about the layout and contents of the button bar, which seems a bit brittle. If it's messier the other way, though, no worries.

andymatuschak (changes requested):
Nice work! Thank you.

packages/ui/src/components/ReviewArea.tsx
growthepie: Yeah, I think that makes more sense as well. Just did it!

andymatuschak (approved):
Looking good! I’m glad you made the change—I think this is cleaner in the end. Please check your Gmail address for a boring logistical question, then this is ready to go.

andymatuschak (changes requested):
Whoops, just noticed that there are a few linter complaints here, if you wouldn't mind fixing these up.

growthepie:
Whoops, forgot to lint. Will push a fix!

growthepie:
Refactored quite a bit to make sure we only use hooks at the top level. Let me know what you think @andymatuschak!

andymatuschak (approved):
Really nice and clean now!

PR #219, opened 2021-06-04
Removing local dependency references from backend

andymatuschak:
Wanted to run this by you @kirkbyo, since it's a bit weird.

`backend` is deployed to Cloud Functions, but the server can't actually access local packages like `@withorbit/api`: they're not actually published to NPM. So our strategy has been to bundle the local dependencies into a single .js which then gets deployed. But to make this strategy work, I have to remove the local dependencies from package.json (otherwise, the Firebase servers try to install them along with all the other dependencies, and fail).

I noticed that you re-added these dependencies recently. Was something breaking when they weren't declared? Everything seems to work appropriately with them gone as far as I can tell, but I wanted to run it past you.


kirkbyo (approved):
Mhmm weird. 

I added the dependencies as part of https://github.com/andymatuschak/orbit/commit/c59ac9f96196361e3c8d2ca4eafaec2e5f254156 which fixed getting `yarn build` to work at the root for fresh clones. I remember for the other packages I needed to explicitly list their local dependencies, but I guess that doesn't apply to the backend?

Anyways, the tests passed in CI so this fix doesn't affect fresh clones. I guess they did not need to be declared in this case. 🤔

andymatuschak:
Cool, thanks for confirming.

Issue #75, opened 2020-09-04
Embedded review areas should update parent iframe width as they're resized


andymatuschak:
Currently the embedded iframes' height is calculated when they're created. But that height should update when the review area's width changes.

PR #222, opened 2021-06-07
Resize ReviewArea web component as the window size changes

kirkbyo:
Closes https://github.com/andymatuschak/orbit/issues/75

**Problem**
As the browser is resized, the `iframe` instantiated by the `OrbitReviewAreaElement` web component does not adjust its height.

**Solution**
I decided to use the `ResizeObserver` API which provides the iframe as the target of the event. Alternatively, a resize window event listener could be attached, but the iframe elements would need to be resolved which could become tricky.

**Demo**
<details>
<summary>GIF of the height changing as the browser is resized</summary>

resize-demo
</details>


andymatuschak (approved):
Nice work! I’ll go ahead and merge.

PR #227, opened 2021-07-09
Adopting modifyEntities in Database, implementing incremental snapshot updating

andymatuschak:
@kirkbyo, I'd be grateful for your eyes on this change, since it involved some plumbing in `IDBDatabaseBackend`.

At a high level, I'm doing two things here (sorry to have combined them into a single commit—it was a bit tangled):
1. Adopting `modifyEntities` in `Database.putEvents`, rather than `getEntities` followed by `putEntities`.
2. Implementing a "fast path" for the common case of updating an entity snapshot, in which the new events can be applied directly on top of the previous snapshot.

To accomplish this, I had to add a "column" to the backends, and I also had to change `modifyEntities`'s API signature to use general types, rather than Dexie-specific types.

While implementing this, I noticed (thanks, failing tests!) that `IDBDatabaseBackend.modifyEntities` actually only supported updating pre-existing entities, but not inserting new ones. I fixed that up too.

### DIFF
--- a/packages/store/src/database.ts
+++ b/packages/store/src/database.ts
@@ -35,9 +35,7 @@ export class Database {
 
   async putEvents(events: Event[]): Promise<void> {
     await this._backend.putEvents(events);
-
-    const affectedEntityIDs = new Set(events.map(({ entityID }) => entityID));
-    await this._updateEntitySnapshots([...affectedEntityIDs]);
+    await this._mergeEventsIntoEntitySnapshots(events);
   }
 
   getEvents<E extends Event, ID extends EventID>(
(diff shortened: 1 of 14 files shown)


andymatuschak:
(ignore the test failures—I've implemented `SQLDatabaseBackend.modifyEntities` in a later commit, but wanted to avoid making this PR too big)

packages/store/src/database.ts
```diff
@@ -68,54 +66,97 @@ export class Database {
...
-    });
-    const lastEventID = events[events.length - 1];
-    if (events.length > 0) {
+    newEvents.sort((a, b) => a.timestampMillis - b.timestampMillis);
+
+    // Most of the time, new events will be ordered after old events (in terms of local client time), which means we can apply them directly on top of our existing snapshot.
+    if (
+      currentRecord &&
+      newEvents.length > 0 &&
+      newEvents[0].timestampMillis > currentRecord.lastEventTimestampMillis
```
kirkbyo, line 121: Is there the assumption that the client will always send their events sorted by timestamp or are we validating that somewhere else in the stack?

packages/store/src/database.ts
```diff
@@ -68,54 +66,97 @@ export class Database {
...
-      predicate: ["entityID", "=", entityID],
-    });
-    const lastEventID = events[events.length - 1];
-    if (events.length > 0) {
+    newEvents.sort((a, b) => a.timestampMillis - b.timestampMillis);
+
+    // Most of the time, new events will be ordered after old events (in terms of local client time), which means we can apply them directly on top of our existing snapshot.
+    if (
+      currentRecord &&
+      newEvents.length > 0 &&
```
kirkbyo, line 120: Should this be validated at the HTTP API level?

packages/store/src/database/backend.ts
```diff
@@ -25,6 +31,9 @@ export interface DatabaseBackend {
 
 // We persist entities wrapped with extra metadata used for updating snapshots.
 export interface DatabaseBackendEntityRecord<E extends Entity> {
-  lastEventID: EventID; // i.e. last in the order as returned by the backend, following logical time, rather than client-local time
   entity: E;
+
+  // Here, "last" refers to the local logical clock (i.e. sequence number), rather than client-local time.
```
kirkbyo, line 36: Took me a few to understand this comment given the addition of `lastEventTimestampMillis` below. Maybe something like:
```
Here, "last" is determined using the local logical (i.e. sequence number), rather than client-local time.
```

packages/store/src/database/backends/indexedDB.ts
```diff
@@ -22,6 +22,18 @@ import {
   DexieEventRowWithPrimaryKey,
 } from "./dexie/tables";
 
+function getEntityRowForEntityRecord(
+  record: DatabaseBackendEntityRecord<Entity>,
+) {
```
kirkbyo, line 27: (feel free to ignore): could we type the return to make it clearer when the Dexie record changes in the future.
```suggestion
function getEntityRowForEntityRecord(
  record: DatabaseBackendEntityRecord<Entity>,
): DexieEntityRow {
```

This way the error will point right to this function rather then to the consumer of it.

packages/store/src/database/backends/indexedDB.ts
```diff
@@ -220,33 +223,39 @@ export class IDBDatabaseBackend implements DatabaseBackend {
...
+          const rowID = entityIDsToRowIDs.get(id);
           transformedRows.push({
-            rowID: row.rowID,
-            id: row.id,
-            entityType: row.entityType,
-            lastEventID: row.lastEventID,
-            data: row.data,
+            ...getEntityRowForEntityRecord(record),
+            // We include the old row ID if we had a record for this row before; otherwise, it's a new entity, and the DB will assign it a row ID.
+            ...(rowID !== undefined && { rowID }),
```
kirkbyo, line 258: Very nice!

kirkbyo (approved):
Looks good; no blockers stood out on my end! Only a couple questions / small nits which feel free to ignore.

packages/store/src/database.ts
andymatuschak: Nope: they're sorted on line 115.

packages/store/src/database.ts
andymatuschak: Practically speaking, `newEvents` can never be empty because we got here by iterating a map which was built by collating events by entity.

But what was in my mind was that a method which recomputes the snapshot for an entity is generally useful, and that we might want to recompute an entity's snapshot in some other circumstance, outside `putEvents`, in which case `newEvents` would be empty, and we'd want to fetch all the events. Arguably, YAGNI… I suppose I was mostly just being defensive.

Issue #172, opened 2021-01-07
Provenance links aren't obviously links


andymatuschak:
Users have commented that they didn't realize the provenance links in reviews are actually links. We should probably add a carat or something to indicate their linkiness.

PR #228, opened 2021-07-12
Make provenance links more obvious

tyranitar:
Takes a crack at https://github.com/andymatuschak/orbit/issues/172. Right now it does the following:
- Add an optional `iconLeftOfTitle` prop to `Button` to show icon next to the title instead of above.
- Use the above to add a link icon to the provenance link button. There is no link icon in the `assets/icons` folder yet, but if this approach seems okay, we can look for an SVG I guess.
- Add `testNotes` and `cache.db` to `.gitignore` for convenience.

Screenshot:
Screen Shot 2021-07-11 at 10 34 48 PM

Putting it up as a draft to get feedback on whether this feels like the right approach.

### DIFF
--- a/packages/ui/src/components/Button.tsx
+++ b/packages/ui/src/components/Button.tsx
@@ -25,6 +25,7 @@ export type ButtonPendingActivationState = "hover" | "pressed" | null;
 type ButtonContents =
   | { title: string; iconName?: IconName }
   | { iconName: IconName; accessibilityLabel: string };
+
 type ButtonAction =
   | {
       onPress: () => void;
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Button.tsx
```diff
@@ -156,32 +159,34 @@ const ButtonInterior = function ButtonImpl(
...
+                {
+                  color: titleColor,
+                },
+              ]}
+              selectable={false}
+              suppressHighlighting={true}
+            >
+              {props.title}
+            </Text>
+          )}
```
tyranitar, line 188: No changes here, just shifted from the wrapping `<div>`.

PR #230, opened 2021-07-28
Add scaffolding for API integration tests

kirkbyo:
Added `packages/api-integration`. It holds the full-stack tests of the backend. For now, this package will only run locally since the tests require an instance of the backend (including Firestore) to be running. We could also put these tests in the `backend` package, but I figured a separate package would be easier to configure how we want the tests to run (or not to run in the case of CI).

I only created basic test cases for the `/actionLogs` route in this PR, but I will add more if you agree with this approach! 

**Note**: I included a few API validation tests in this PR. Although this portion of the code is handled by `Ajv`, I see these tests as a contract to ensure no regressions down the line (instead of testing the internal `Ajv`). However, I could also see why we wouldn't want to pollute the code with validation tests, so let me know if you have any preferences!


packages/api-integration/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../tsconfig.base",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "lib": ["es6", "dom"],
```
andymatuschak, line 6: Don't think we need this line? We need it in packages which need the web API types.

packages/api-integration/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "@types/node": "^16.4.3",
+    "babel-jest": "^27.0.6",
+    "jest": "^27.0.6",
+    "typescript": "^4.3.5"
+  },
+  "dependencies": {
+    "isomorphic-fetch": "^3.0.0"
+  },
+  "scripts": {
+    "test": "HOST=http://localhost:5001/metabook-system/us-central1 jest"
```
andymatuschak, line 19: If you think it'd be cleaner, the `backend` tests have some code which actually starts up and tears down the Firebase emulator within the test execution: https://github.com/andymatuschak/orbit/blob/master/packages/backend/src/backend/firebaseSupport/__tests__/firebaseTesting.ts. Though if you wanted to use that you might have to move this into the `backend` package.

andymatuschak (changes requested):
Nice work! Broad approach looks good, and I feel safer already seeing these kinds of tests get added. I could go either way on the question of whether this should be a separate package or in backend; happy to let you make the call.

andymatuschak (approved):
Whoops—er, I didn't mean to click "request changes." Feel free to land at will. :

kirkbyo:
@andymatuschak Thank you for pointing me towards the `firebaseSupport` helpers, I had no idea those existed and that we were already spinning up the Firebase emulator in CI; neat! 

Given that this infrastructure is already in the backend, I decided to move the tests to `package/backend/integration` and run them with the backend tests. This comes with a trade off that the tests need to be ran sequentially, [`--runInBand`](https://jestjs.io/docs/cli#--runinband). Without this flag, the test become flaky due to running `firestoreDatabaseBackend.test.ts` and the integration tests due to `firestoreDatabaseBackend.test.ts` wiping the DB after each test (it was very flaky locally). 

This works for now, but I suspect the test times might become too slow in the future, so at that point we could break out the integration tests into its own package and maybe store the `firebaseSupport` from within there? Not sure. Let me know if you have any issues with this approach!

packages/backend/package.json
```diff
@@ -13,7 +13,7 @@
     "deployCORS": "gsutil cors set ../backend/firebase-storage-cors.json gs://metabook-system.appspot.com",
     "deployFirestoreConfig": "firebase deploy --only firestore",
     "fetchFirestoreIndexes": "firebase firestore:indexes > ../backend/firestore.indexes.json",
-    "test": "jest"
+    "test": "jest --runInBand"
```
andymatuschak, line 16: Ah: I'd actually meant to add this a few days ago anyway! My new Firestore `DatabaseBackend` test's implementations need this anyway. Great!

andymatuschak (approved):
Nice: now that you've moved these into the `backend` package, this approach seems cleaner. I'm not too worried about CI runtime yet, but we may need to deal with that later!

I imagine that everything's working as expected, but I'm a little surprised that you're able to put TypeScript sources in `backend/integration` (as opposed to `backend/src/integration`) because `backend/tsconfig.json` only includes `src/**/*`, which doesn't match these tests. I guess maybe the Jest TS integration doesn't care! Assuming everything works as it's supposed to, that's fine!

kirkbyo:
@andymatuschak I was a bit surprised as well, I expected some odd compilation errors, but it works locally on my end and it ran successfully in CI. I suspect it has something todo with the Babel config ignoring the included sources defined in the `tsconfig`. Not sure 🤔.

I will go ahead and merge and if we run into any issues, I will move the files into `backend/src`

PR #231, opened 2021-07-31
Adding initial implementation of sync API

andymatuschak:
@kirkbyo, I was hoping I could get your eyes on this not only because sync is error-prone but also because I ended up making some changes to the database implementations.

In particular, I've made `putEvents` ignore events that are already stored. My reasoning: in a networked system, you need to be able to retry operations safely—ideally idempotently. The particularly tough situation is when a PUT succeeds and the new data is committed to the DB, but the connection drops before the client gets the 200. In this case, there's no way for the client to know that the events have gone through. They have to send them again. And that has to be OK with the server! I think the same logic mostly applies in local circumstances. The alternative is to force clients to first make sure that events aren't yet stored, e.g. by getting them, but that path leaves clients open to race conditions.

Note that as of this PR, we currently always performs a full sync, since I haven't yet added support for reading/writing the bookkeeping metadata to the stores, but we should get incremental sync "for free" once I've done that.

I separated the key "jobs" of this PR into a couple different commits for your convenience.

### DIFF
--- a/packages/store-web/src/indexedDB.ts
+++ b/packages/store-web/src/indexedDB.ts
@@ -1,4 +1,5 @@
 import Dexie, {
+  BulkError,
   Collection,
   IndexableType,
   PromiseExtended,
@@ -258,15 +259,47 @@ export class IDBDatabaseBackend implements DatabaseBackend {
   }
 
   async putEvents(events: Event[]): Promise<void> {
-    this.db.transaction("rw", this.db.events, async () => {
-      await this.db.events.bulkAdd(
-        events.map((event) => ({
-          entityID: event.entityID,
-          id: event.id,
-          data: JSON.stringify(event),
-        })),
-      );
-    });
+    await this.db
+      .transaction("rw", this.db.events, async () => {
+        await this.db.events
+          .bulkAdd(
+            events.map((event) => ({
+              entityID: event.entityID,
+              id: event.id,
+              data: JSON.stringify(event),
+            })),
+          )
+          .catch("BulkError", (error: BulkError) => {
+            // We'll ignore attempts to add duplicate events.
+            // If there are any errors which aren't duplicates, though, we'll just fail.
+            if (
+              Object.values(error.failures).some(
+                (failure) => failure.name !== Dexie.errnames.Constraint,
+              )
+            ) {
+              throw error;
+            }
+
+            // We'll find all events which *aren't* duplicate, then attempt to write those.
+            const duplicateIndices = new Set(
+              Object.keys(error.failures).map((key) => Number.parseInt(key)),
+            );
+            // We have to wait until the transaction finishes aborting, then attempt another.
+            throw {
+              name: "__retry",
+              validEvents: events.filter((e, i) => !duplicateIndices.has(i)),
+            };
+          });
+      })
+      .catch(async (error) => {
+        if (error.name === "__retry") {
+          if (error.validEvents) {
+            await this.putEvents(error.validEvents);
+          }
+        } else {
+          throw error;
+        }
+      });
   }
 
   async _fetchPrimaryKeyFromUniqueKey<Row, PK>(
--- a/packages/sync/src/__tests__/setup.ts
+++ b/packages/sync/src/__tests__/setup.ts
@@ -1,5 +1,0 @@
-import fetchMock from "jest-fetch-mock";
-fetchMock.enableMocks();
-
-// Disable by default; we'll enable within tests, where needed.
-fetchMock.dontMock();
--- a/packages/sync/src/__tests__/setup.ts
+++ b/packages/sync/src/__tests__/setup.ts
@@ -1,0 +1,5 @@
+import fetchMock from "jest-fetch-mock";
+fetchMock.enableMocks();
+
+// Disable by default; we'll enable within tests, where needed.
+fetchMock.dontMock();
--- a/packages/sync/src/orbitStoreSyncInterface.ts
+++ b/packages/sync/src/orbitStoreSyncInterface.ts
@@ -1,0 +1,59 @@
+import { OrbitStore } from "@withorbit/store-shared";
+import {
+  AttachmentID,
+  AttachmentMIMEType,
+  Event,
+  EventID,
+} from "@withorbit/core2";
+import { SyncInterface } from "./syncInterface";
+
+export class OrbitStoreSyncInterface implements SyncInterface {
+  id: string;
+  private readonly _orbitStore: OrbitStore;
+
+  constructor(orbitStore: OrbitStore, syncID: string) {
+    this.id = syncID;
+    this._orbitStore = orbitStore;
+  }
+
+  async putEvents(events: Event[]): Promise<void> {
+    await this._orbitStore.database.putEvents(events);
+  }
+
+  async listEvents(
+    afterEventID: EventID | null,
+    limit: number,
+  ): Promise<Event[]> {
+    return await this._orbitStore.database.listEvents({
+      afterID: afterEventID ?? undefined,
+      limit,
+    });
+  }
+
+  async putAttachment(
+    sourceURL: string,
+    id: AttachmentID,
+    type: AttachmentMIMEType,
+  ): Promise<void> {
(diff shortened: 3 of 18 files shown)


packages/sync/src/orbitStoreSyncInterface.ts
```diff
@@ -0,0 +1,59 @@
+import { OrbitStore } from "@withorbit/store-shared";
+import {
+  AttachmentID,
+  AttachmentMIMEType,
+  Event,
+  EventID,
+} from "@withorbit/core2";
+import { SyncInterface } from "./syncInterface";
+
+export class OrbitStoreSyncInterface implements SyncInterface {
```
kirkbyo, line 10: nit: `OrbitStoreSyncManager` instead maybe? Wouldn't `Interface` in the name imply an abstract implementation?

packages/sync/src/sync.ts
```diff
@@ -0,0 +1,77 @@
...
+      await destination.putEvents(events);
+      afterEventID = events[events.length - 1].id;
+      await setLatestSentEventID(afterEventID);
+    } else {
+      break;
+    }
+  }
+}
+
+const receiveBatchSize = 500;
```
kirkbyo, line 76: nit: I believe I have seen this convention elsewhere in the code?
```suggestion
const RECEIVE_BATCH_SIZE = 500;
```

packages/sync/src/syncInterface.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  AttachmentID,
+  AttachmentMIMEType,
+  Event,
+  EventID,
+} from "@withorbit/core2";
+
+export interface SyncInterface {
+  // This ID is used to track state associated with the remote destination associated with this interface. Different destinations (e.g. test servers, production server, local host) should have different IDs.
+  id: string;
```
kirkbyo, line 10: Is this used for logging purposes or do you expect some utility from this metadata in the future?

packages/store-web/src/indexedDB.ts
```diff
@@ -258,15 +259,47 @@ export class IDBDatabaseBackend implements DatabaseBackend {
...
+              name: "__retry",
+              validEvents: events.filter((e, i) => !duplicateIndices.has(i)),
+            };
+          });
+      })
+      .catch(async (error) => {
+        if (error.name === "__retry") {
+          if (error.validEvents) {
+            await this.putEvents(error.validEvents);
+          }
```
kirkbyo, line 298: Should we include some logging for this?

kirkbyo (approved):
Nothing major stood out on my end; just a few small nits (feel free to ignore)! 

One thing though, the attachment sync API isn't clear to me right now. A few questions:
- Is an `AttachmentID stored within the `Events`?
- If a user updates an attachment, is this a new attachment or would the same ID be reused?
- When are the attachments persisted? Would this happen before or after events?

packages/sync/src/__tests__/setup.ts
```diff
@@ -0,0 +1,5 @@
+import fetchMock from "jest-fetch-mock";
+fetchMock.enableMocks();
+
+// Disable by default; we'll enable within tests, where needed.
+fetchMock.dontMock();
```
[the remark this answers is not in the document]
andymatuschak: Nope, not strictly needed. It's just hygienic: when you add `jest-fetch-mock`, it's on by default. But global monkey-patching side effects freak me out. I want it to only affect tests where we clearly opt into mocking `fetch` in that file, as opposed to having it silently/secretly affect all tests. If some code is unexpectedly `fetch()`ing, the mocking might obscure that.

packages/sync/src/orbitStoreSyncInterface.ts
andymatuschak: Good point—in TypeScript, I suppose it would. Perils of working in so many languages…

This is really an "adapter" in the Gang of Four sense: the idea is that you can sync from/to several kinds of things (the server via an `APIClient`, another local `OrbitStore`, etc). I wouldn't want to call it a "manager" because it doesn't do anything itself; it just translates requests to something else which does the work.

packages/sync/src/sync.ts
andymatuschak: Perhaps! I do prefer to stick to camel case for variable identifiers, even if they're magic number constants. I may have slipped up somewhere, or you may be thinking of another project: shouting snake case is quite common for constants like this in the JS world.

packages/sync/src/syncInterface.ts
andymatuschak: We use it when storing the last event ID sent/received to/from the sync destination. An ID is necessary to keep different state for different sync destinations (e.g. perhaps you're syncing to both a local server and the production server).

See the next PR for how it's used:
https://github.com/andymatuschak/orbit/pull/232/files#diff-5b6b6441fbdb9546805cc316cbf46bd767f35dd9e60db2897beab1d462b561c3R17

I improved the comment here.

packages/store-web/src/indexedDB.ts
andymatuschak: Ah, so—getting here actually doesn't constitute an error condition! It just means we were asked to insert some events which already exist, and we're ignoring those which already exist. This will actually happen every time we sync.

andymatuschak:
Thanks for the careful read, @kirkbyo! Let's make sure the attachment bits make sense to you before I land this.

> Is an AttachmentID stored within Events?

There are currently two types of entities: `Task` and `AttachmentReference`. The latter is created when a client emits an `AttachmentIngestEvent`, which (like all events) includes its referenced entity ID (in this case an `AttachmentID`).

More concretely: if you're implementing an Orbit client, and you want to add a task which involves an attachment, you'd do this:

1. Add the attachment binary contents to your `AttachmentStore` under some attachment ID you generate.
2. Add an `AttachmentIngestEvent` to your database referencing the ID from step 1.
3. Add a `TaskIngestEvent` referencing the ID from step 1.

> If a user updates an attachment, is this a new attachment or would the same ID be reused?

The correct semantic probably depends on exactly how this is expressed in the UI. It could either mean creating a new attachment and updating the task contents to point to that new ID, or emitting something like an `AttachmentContentChange` event (no such event exists yet, and it would be a bit tricky since I don't want to store blobs in the database… would probably involve modeling "versions" of attachments explicitly).

> When are the attachments sent? Would this happen before or after the events sync?

See this bit of the sync routine:
https://github.com/andymatuschak/orbit/pull/231/files#diff-5b6b6441fbdb9546805cc316cbf46bd767f35dd9e60db2897beab1d462b561c3R64

For each batch of events sent/received, the sync implementation looks through them for `AttachmentIngestEvent`s. It then sends/receives the associated attachment(s). Only once it's done that will it actually store/transmit the events referencing the attachments.

kirkbyo:
Thank you for the clarifications @andymatuschak, it all makes sense to me! I am really liking the abstractions between syncing, communication layer, and the different backends!

---- 
Notes on some of the edge cases I was thinking about:

> See this bit of the sync routine:
https://github.com/andymatuschak/orbit/pull/231/files#diff-5b6b6441fbdb9546805cc316cbf46bd767f35dd9e60db2897beab1d462b561c3R64

Okay perfect, the attachments are pushed before the events. I was worried about potentially eager clients requesting an asset before the attachment itself has persisted to the backend and resulting in some weird errors on the client.

> The correct semantic probably depends on exactly how this is expressed in the UI. It could either mean creating a new attachment and updating the task contents to point to that new ID, or emitting something like an AttachmentContentChange event (no such event exists yet, and it would be a bit tricky since I don't want to store blobs in the database… would probably involve modeling "versions" of attachments explicitly)

Gotcha, makes sense! I was worried about the case where an older attachment (that was set locally and offline) overriding a newer attachment due to it being synced last. However, since the attachment IDs are stored within the events, the last event with the latest attachment ID would be used 👍

kirkbyo (approved):

andymatuschak:
Thanks for looking. The abstraction boundaries are feeling good to me now, too. A big improvement in solidity and flexibility on the previous iteration—hopefully will justify all the sweat involved in this restructuring. :)

PR #233, opened 2021-08-05
Migrate packages/ui to core2

kirkbyo:
**How was this tested?**
- Manually took screenshots of components and compared them to the current version to ensure there were no visual regressions. 
    - _Idle wondering_: It might be worth it to setup screenshot testing locally for this package before landing this PR.
- Ensured that none of the storybook stories were broken.

**Note on build errors**:
- All the errors are due to `packages/app` using the `core` types right now. Once those changes are made, this PR can land in tandem. `packages/ui` compiles locally right now!


packages/sample-data/src/core2/testClozeSpec.ts
```diff
@@ -0,0 +1,39 @@
+import {
+  ClozeTaskContent,
+  MemoryTaskSpec,
+  TaskContentType,
+  TaskSpecType,
+} from "@withorbit/core2";
+
+export const testClozeSpec: MemoryTaskSpec<ClozeTaskContent> = {
```
kirkbyo, line 8: _PS_: I moved this out of `sample-data/src/core2/testQASpec.ts` to match the layout of the sample core1 types (i.e. `sample-data/src/testClozePrompt.ts`).

packages/ui/src/components/Card.tsx
```diff
@@ -33,23 +26,17 @@ import {
...
+function getQAPromptContents<TC extends QATaskContent | ClozeTaskContent>(
+  reviewItem: ReviewAreaItem<TC>,
+): QATaskContent {
+  switch (reviewItem.spec.content.type) {
+    case TaskContentType.QA:
+      // This cast should not be needed... My IDE does not throw any error
+      // without the cast, but TSC throws:
+      // `Type 'TC' is not assignable to type 'QATaskContent'`.
+      // casting resolves the type issue...
+      return reviewItem.spec.content as QATaskContent;
```
kirkbyo, line 38: Very weird error... Could this be due to my IDE and TSC using different versions of typescript? Not sure how to justify this error otherwise.

packages/ui/src/components/Card.tsx
```diff
@@ -122,115 +109,80 @@ function useAnimatingStyles(backIsRevealed: boolean): {
...
+        "{" +
+        mutatingText.slice(range.startIndex, range.startIndex + range.length) +
+        "}" +
+        mutatingText.slice(range.startIndex + range.length);
     } else {
-      output += match[1]; // strip the braces
+      mutatingText =
+        mutatingText.slice(0, range.startIndex) +
+        clozeBlankSentinel +
+        mutatingText.slice(range.startIndex + range.length);
```
kirkbyo, line 174: Ended up not handling the hint case. I will create an issue for that when this PR lands.

packages/ui/src/styles/colors.ts
```diff
@@ -100,5 +100,6 @@ const _palettes = {
   },
 };
 
+const typedKeys = Object.keys as <T>(o: T) => Extract<keyof T, string>[];
```
kirkbyo, line 103: _FYI_: I decided to inline due to the only other usage of this function is within `packages/core`. That being said, it could be moved to `core2` if you expect other usages!

packages/ui/src/components/Card.tsx
andymatuschak: Ah, I think this is because TS can't do this kind of type narrowing inference across multiple projections. To put it another way, this should work:

```
const {content} = reviewItem.spec;
switch (content.type) {
  case TaskContentType.QA:
    return content;
```

packages/ui/src/components/Card.tsx
```diff
@@ -122,115 +109,80 @@ function useAnimatingStyles(backIsRevealed: boolean): {
...
+          ellipsizeMode="tail"
+        >
+          {provenance.title}
+        </Text>
+      )}
+    </View>
+  ) : null;
 }
 
 function formatClozePromptContents(
```
andymatuschak, line 156: Gosh this whole thing is much better now.

packages/ui/src/components/StarburstLegend.tsx
```diff
@@ -97,7 +98,23 @@ export interface StarburstLegendProps {
...
+  for (let i = 0; i < maxIntervals; i++) {
+    sequence.push({
+      interval:
+        sequence[sequence.length - 1].interval * config.intervalGrowthFactor,
+      label: "tmp",
+    });
+  }
+  return sequence;
+}
+
```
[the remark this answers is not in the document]
andymatuschak: Right! I think the right way to think about it is that this is now a `ui` concern: at least for the moment, it's really just presentational. But I did find that it's best to massage the labels a little bit, so that they aren't ugly "18 days" or whatever: I think we probably do still want to hard-code a literal sequence so that we can have "nice" labels as before.

andymatuschak (changes requested):
Wow, outstanding! This is great, and in a number of places made the code better.

Would you like to handle the legend labels in this PR or in a follow-on? Can land this in a branch until `app` is fully ready.

packages/ui/src/components/StarburstLegend.tsx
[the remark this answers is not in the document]
kirkbyo: Gotcha! I just implemented a fixed interval sequence in the latest commit, let me know if that's what you had in mind!

packages/ui/src/components/Card.tsx
kirkbyo: Darn, I was hoping this would work. Looks, like the issue is still persisting 😕

Screen Shot 2021-08-06 at 12 00 05 PM

andymatuschak:
Closing since I've merged into `core2-app`

PR #234, opened 2021-08-05
Fixed race condition in Dexie hooks

andymatuschak:
Dexie's hooks aren't asynchronous, so we can't actually async/await in them as we were doing in our hook implementations. This happened to work in fakeindexeddb, but the timing of the browser's implementation was different in a way which manifested subtle race conditions. I've moved the hook implementations into the transaction, where we can actually `await` on them.


packages/store-web/src/dexie/dexie.ts
```diff
@@ -51,72 +51,5 @@ export class DexieDatabase extends Dexie {
...
-  transaction: Transaction,
-) {
-  if (obj.entityType === EntityType.Task) {
-    await deleteAllDerivedTaskComponentsByTaskID(obj.id as TaskID, transaction);
-
-    insertDerivedTaskComponent({ ...obj, ...modifications }, transaction);
   }
 }
-
-async function onEntitiesDelete(
```
andymatuschak, line 84: n.b. there's no equivalent in my patch; entities can never be deleted once created (would violate our event sourcing semantics)—instead they can have an isDeleted flag or whatever

kirkbyo (approved):
Ah shucks; sorry about that! 

Everything looks good!

PR #235, opened 2021-08-15
Fix mismatch between the docs code snippet and the prompts that are displayed

ZanderKasanjian:
Minor update to the HTML code snippet on the docs to make it match the prompts that are displayed.


andymatuschak:
Helpful; thank you, Zander!

Issue #236, opened 2021-08-19
ajv validation is incorrectly permitting extra properties on Events


andymatuschak:
See test in `events.test.ts`: "[PATCH] validation/it fails when extra properties are provided".

In this test, we modify an incoming `Event` by adding an extra property. The API should 400, but instead it succeeds.

The JSON schema looks correct as far as I can tell: `definitions/TaskIngestEvent` does indeed include `additionalProperties: false`. The input fails to validate as expected [in this online sandbox](https://www.jsonschemavalidator.net/s/L0Th35Bm).

This may be an AJV bug; I'm not sure.

andymatuschak:
@kirkbyo, would you mind taking a look at this? Hoping you might see some obvious issue.

PR #241, opened 2021-08-20
Fix AJV validation is incorrectly permitting extra properties on Events

kirkbyo:
Closes https://github.com/andymatuschak/orbit/issues/236

**Problem**: When including an extra property like in the case of the [PATCH `/api/events`](https://github.com/andymatuschak/orbit/blob/master/packages/backend/src/__tests__/integration/events.test.ts#L92) route, the API validator will incorrectly succeed the route despite this property being an obvious validation error.

**Solution**: I believe in the process of updating dependencies recently (maybe https://github.com/andymatuschak/orbit/commit/4eb0828b5ad3b73f3242f436334027347cbc2ab3?), the AJV version upgraded. With newer versions of AJV, a new property `instancePath` is included and the `schemaPath` seems to behave slightly different for nested objects. To fix our issue, we just need to identify that the `instancePath` is the root (i.e. an empty string). 

**Some idle notes**: I have never been too keen on my solution to this unsupported route problem given how fragile it is. One alternative is to identify the routes from the JSON schema (`Object.keys(schema.definitions)`) and then check against this list when trying to determine if the route is supported. That being said, this solution quickly grows in complexity when trying to handle dynamic routes like `/attachments/:id`.

### DIFF
--- a/packages/api/src/validation/ajvAPIValidator.ts
+++ b/packages/api/src/validation/ajvAPIValidator.ts
@@ -85,7 +85,10 @@ export class AjvAPIValidator<T extends AjvSchema> implements APIValidator {
 
   private _isUnsupportedRoute(): boolean {
     if (this.validator.errors) {
-      return this.validator.errors[0].schemaPath === "#/additionalProperties";
+      return (
+        this.validator.errors[0].instancePath === "" &&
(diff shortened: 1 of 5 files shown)


andymatuschak (approved):
Ah, I see! Thank you for fixing (and thank you for fixing my lint errors too).

Issue #240, opened 2021-08-19
Add validation to Database


andymatuschak:
Thanks to @kirkbyo, we validate client data at our API endpoints. But our clients should really validate data locally, too. Otherwise, a buggy script which inserts faulty data into a local Orbit store will leave the store un-syncable.

PR #244, opened 2021-09-12
Validate event schema on putEvents

kirkbyo:
Closes https://github.com/andymatuschak/orbit/issues/240

**What does this do?**
Validates at runtime that the `events` parameter on `database.putEvents` has the correct schema.

**Why?**
SQLite and IndexedDB do not validate the schema before any mutation. Thus, if the schema is incorrect, the users local database might move into a bad state and making it impossible to sync.

**How?**
Using a similar validation strategy as the API, a JSON schema was generated for the `Event[]` type. This schema is used to verify that the `events` parameter of `putEvents` is valid at runtime (thanks to `Ajv`)

### DIFF
--- a/packages/store-shared/package.json
+++ b/packages/store-shared/package.json
@@ -9,15 +9,18 @@
   "exports": "./dist/index.js",
   "sideEffects": false,
   "scripts": {
-    "build": "tsc -b"
+    "build": "tsc -b",
+    "generateSchema": "typescript-json-schema src/validation/eventsValidator.ts Events -o src/validation/events.json --noExtraProps --required --ignoreErrors --strictNullChecks"
   },
   "dependencies": {
     "@withorbit/core": "0.0.1",
+    "ajv": "8.6.2",
     "base64-js": "^1.5.1"
   },
   "devDependencies": {
     "@types/base64-js": "^1.3.0",
     "@types/jest": "^26.0.24",
-    "typescript": "^4.2.4"
+    "typescript": "^4.2.4",
+    "typescript-json-schema": "^0.50.1"
   }
 }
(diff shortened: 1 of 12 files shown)


packages/store-shared/package.json
```diff
@@ -9,15 +9,18 @@
   "exports": "./dist/index.js",
   "sideEffects": false,
   "scripts": {
-    "build": "tsc -b"
+    "build": "tsc -b",
+    "generateSchema": "typescript-json-schema src/validation/eventsValidator.ts Events -o src/validation/events.json --noExtraProps --required --ignoreErrors --strictNullChecks"
   },
   "dependencies": {
     "@withorbit/core": "0.0.1",
+    "ajv": "8.6.2",
```
kirkbyo, line 17: note: I had to manually specify this version (as opposed to `^8.6.2`) to avoid `store-fs` using the `ajv` hosted at the project root which is using version 6 of the package.

packages/store-shared/src/validation/AjvEventsValidator.ts
```diff
@@ -0,0 +1,35 @@
...
+            message: `${error.schemaPath} ${error.message}`,
+          })),
+        };
+      } else {
+        return {
+          errors: [{ message: "An unknown validation error occurred" }],
+        };
+      }
+    }
+    // TODO: validate the existence of IDs
```
kirkbyo, line 32: I left this is as a todo, but would be happy to address it in this PR! I just wasn't sure which IDs would make sense to have their existence validated? Right now, if you try to reschedule before the initial ingest, it will fail. What other cases should we protect against?

andymatuschak: Yeah, this is a key limitation. Some portion of the validity of events is contingent on the state of the entity it's being applied to.

But I think that's OK: the `Database` will still fail `putEvents` when trying to apply the event to the entity.

andymatuschak (approved):
Looks great, Ozzie. I'll go ahead and merge this in!

Issue #224, opened 2021-06-11
Add hints to review buttons for keyboard shortcuts


andymatuschak:
In #218, @growthepie kindly added support for keyboard shortcuts to our review buttons. Now we need to tell users about them!

I don't have any immediate opinions on the right design for this. We're a bit constrained on those buttons in terms of real estate. A tooltip on hover after a moment of dwell might be a reasonable place to start.

PR #247, opened 2021-11-08
ui/Tooltip - basic implementation of tooltip overlay

Goondrious:
I did a basic implementation for #224 and tried to keep it flexible, since there hadn't been a design discussion yet. I tried to look at some existing packages, but they didn't seem compatible with the Hoverable implementation.

I'm pretty new to typescript, react native and this project, so I'd be happy to refactor to conform to any standards that I missed.

Screen Shot 2021-11-08 at 5 26 31 PM
Screen Shot 2021-11-08 at 5 27 02 PM
Screen Shot 2021-11-08 at 5 26 55 PM
Screen Shot 2021-11-08 at 5 26 45 PM

Immediate TODOs or improvements would be:
- [] more configurable style related props (e.g. color of background & text)
- [] stay on screen
- [] switch the usage from supplying a `ref` of the desired component to composition with built-in hover/touch/arbitrary trigger

### DIFF
--- a/packages/ui/src/components/Tooltip.stories.tsx
+++ b/packages/ui/src/components/Tooltip.stories.tsx
@@ -1,0 +1,93 @@
+import { boolean, number, text, select, withKnobs } from "@storybook/addon-knobs";
+
+import React, { useState, useRef } from "react";
+import Hoverable from "./Hoverable";
+import Tooltip from "./Tooltip";
+import { Pressable, View } from "react-native";
+import { MutableRefObject } from "react";
+
+export default {
+  title: "Tooltip",
+  component: Tooltip,
+  decorators: [withKnobs]
+};
+
+export function Text() {
+  const [isHovered, setHovered] = useState(false);
+  const ref = useRef<View | null>(null);
+  return (
+    <Hoverable
+      onHoverIn={() => setHovered(true)}
+      onHoverOut={() => setHovered(false)}
+    >
+      <View
+        ref={ref}
+        style={{ border: "1px solid black", width: "200px", height: "100px", padding: 20, margin: 100 }}
+      >
+        <Tooltip 
+          show={isHovered}
+          placement={select("direction", ["top", "right", "bottom", "left", "overlay", "inset-bottom-left", "inset-bottom-right", "inset-top-left", "inset-top-right", "cursor"], "top")}
+          text="Tooltip text"
+          anchorRef={ref}
+        />
+        <div>Some Text That Is Important</div>
+      </View>
+    </Hoverable>
+  );
+}
+
+export function Children() {
+  const [isHovered, setHovered] = useState(false);
+  const ref = useRef<View | null>(null);
+  return (
+    <Hoverable
+      onHoverIn={() => setHovered(true)}
+      onHoverOut={() => setHovered(false)}
+    >
+      <View
+        ref={ref}
+        style={{ border: "1px solid black", width: "200px", height: "100px", padding: 20, margin: 100 }}
+      >
+        <Tooltip 
+          show={isHovered}
+          placement={select("placement", ["top", "right", "bottom", "left", "overlay", "inset-bottom-left", "inset-bottom-right", "inset-top-left", "inset-top-right", "cursor"], "top")}
+          text="Tooltip text"
+          anchorRef={ref}
+        >
+          <View style={{ flexDirection: "row", width: "100%", height: "100%", alignItems: "center", justifyContent: "space-around" }}>
+            <View style={{ border: "1px solid red", backgroundColor: "red" }}>
+              <p>L</p>
+            </View>
+            <View style={{ border: "1px solid blue", backgroundColor: "blue" }}>
+              <p>R</p>
+            </View>
+          </View>
+        </Tooltip>
+        <div>Some Text That Is Important</div>
+      </View>
+    </Hoverable>
+  );
+}
+
+export function Button() {
+  const [isHovered, setHovered] = useState(false);
+  const ref = useRef<View | null>(null);
+  return (
+    <Hoverable
+      onHoverIn={() => setHovered(true)}
+      onHoverOut={() => setHovered(false)}
+    >
+      <Pressable
+        style={{ border: "1px solid black", width: "120px", height: "50px", margin: 150 }}
+        ref={ref}>
+        <p>A Pressable</p>
+        <Tooltip 
+          show={isHovered}
+          placement={select("placement", ["top", "right", "bottom", "left", "overlay", "inset-bottom-left", "inset-bottom-right", "inset-top-left", "inset-top-right", "cursor"], "top")}
+          text="Tooltip text"
+          anchorRef={ref}
+        />
+      </Pressable>
+    </Hoverable>
+  );
+}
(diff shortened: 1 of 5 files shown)


andymatuschak:
(I'm out of town this week but will get back to you next week—thank you!)

packages/ui/src/components/Tooltip.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const ref = useRef<View | null>(null);
+  return (
+    <Hoverable
+      onHoverIn={() => setHovered(true)}
+      onHoverOut={() => setHovered(false)}
+    >
+      <View
+        ref={ref}
+        style={{
+          borderWidth: 1,
```
andymatuschak, line 25: Just FYI: the `border` shorthand property doesn't exist in React Native (adds unnecessary parsing overhead… and breaks the type checker! :) )

packages/ui/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,332 @@
...
+} from "react";
+import {
+  ColorValue,
+  StyleProp,
+  StyleSheet,
+  View,
+  ViewStyle,
+} from "react-native";
+
+const styles = StyleSheet.create({
```
andymatuschak, line 17: You did the right thing by just taking a stab here. I'll come back through and match the art direction once we get everything squared away.

packages/ui/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,332 @@
...
+  | "inset-top-left"
+  | "inset-top-right"
+  | "cursor";
+
+export type TooltipProps = {
+  placement?: TooltipPosition;
+  stayOnScreen?: boolean;
+  show: boolean;
+  anchorRef: React.MutableRefObject<View | null>;
+  children: React.ReactNode;
```
andymatuschak, line 96: Hope you don't mind: I modified this from being `text | children` to just being `children`, since `children` can _be_ a string in React (e.g. `<Foo>bar</Foo>`). We can make the union type work if you prefer, but it'll require some extra assertions in the component implementation to pass the type checks.

packages/ui/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,332 @@
...
+  width?: number;
+  height?: number;
+};
+
+type TooltipState = {
+  tooltipPosition?: UIPosition;
+  trianglePosition?: UIPosition;
+};
+
+type Position = {
```
andymatuschak, line 196: I had to replace `Measure` with these `Position` and `Size` types: many of the errors in this file were due to the approach of having every property be optional. TS couldn't prove that the fields would actually exist at runtime as you expected.

packages/ui/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,332 @@
...
+export default (props: TooltipProps) => {
+  const { placement = "top", show, anchorRef, children } = props;
+
+  const tooltipRef = useRef<View | null>(null);
+  const containerRef = useRef<View | null>(null);
+
+  const [{ tooltipPosition, trianglePosition }, dispatch] = useReducer(
+    (
+      state: TooltipState,
+      { type, payload }: { type: string; payload: any },
```
andymatuschak, line 215: Ah, this `any`! Exactly what I try to avoid by using TypeScript. The trouble here is that if I change the argument to `getPositions` (e.g. I thought I might rename `anchor` to `anchorSize` to be clearer), now this code will fail at runtime but won't generate any errors at build time! That's no good.

One alternative is to specify the types specifically: `{type: "init", payload: {placement: string, /* etc */}} | {type: "move", payload: Position}`. But rather than make the change myself, I thought I'd ask: might it be more straightforward to inline the state updates into the effects below? Each command is only used from one place, so I'm not sure `useReducer` is buying us anything here, and it costs us locality. I'm OK either way, but I do want to ditch the `any`!

packages/ui/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,332 @@
...
+  }, [anchorMeasure, show, placement]);
+
+  const trackCursor = useCallback(
+    (e: React.MouseEvent) => {
+      if (!containerPosition) return;
+      dispatch({
+        type: "move",
+        payload: {
+          top: e.clientY - containerPosition.y,
+          left: e.clientX - containerPosition.x,
```
andymatuschak, line 278: Hm, aren't we calculating the difference between a viewport coordinate and a document coordinate here? Isn't this only correct if the document is not scrolled?

packages/ui/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,332 @@
...
+      });
+    },
+    [containerPosition],
+  );
+
+  const mouseEvents = useMemo(
+    () =>
+      placement === "cursor"
+        ? {
+            onMouseMove: trackCursor,
```
andymatuschak, line 289: Bad news here: `onMouseMove` isn't available in React Native! It happens to work in the Storybook because React Native Web implements it, but it won't work in the desktop app. Looks like we can use [react-native-hoverable](https://github.com/ryanlntn/react-native-hoverable). But before we go down that road, let me suggest: I think this is probably unnecessary! I'm fine—at least for now—with anchoring the tooltip to the container view based on its fixed position, rather than having the tooltip follow the mouse. Would you be okay with that simpler approach?

andymatuschak (changes requested):
Oh gosh, I'm terribly sorry that I lost track of this, @Goondrious. I took some time today to run through your changes.

You mentioned that this is your first attempt using TypeScript: an ambitious PR! :) Unfortunately, the TypeScript compiler had many complaints about the types in this patch. You can see them by running `yarn build` in the relevant module's subpath—or by using your editor's TypeScript integration. I went ahead and fixed most of the type errors, which involved changing the structure a bit in a few places. A few issues remain which require some discussion—see inline comments.

I also fixed a number of linter issues. That's my fault—there's no developer documentation for this project yet, so you had no way to know that existed!—but you can run lint checks with `yarn lint` in the root of the repository. Many of them can be automatically fixed with `yarn lint:fix`.

Thanks again for this patch and for your patience!

Goondrious:
No worries @andymatuschak , thanks for the response.

My TypeScript experience has just been in passing, though now I've used it more in personal projects. I feel bad that you went through and did some of the basic tidying up, which I had purposefully left assuming that much would change through any design discussion. Sorry if it caused unnecessary effort/stress on your part. Thank you! I'll review your changes, as I'm sure it'll be a good learning experience. I was actually anticipating a blanket, "please fix type/lint errors".

Hoping to go through more specific comments later this week.

Goondrious:
Apologies, I got caught up with some other work and holidays. Hopefully back to this soon.

packages/ui/src/components/Tooltip.tsx
Goondrious: Yeah, makes sense. I think I had in mind that the flexibility of passing children or a "simpler" `text=...` would be nice, but not tied to it.

packages/ui/src/components/Tooltip.tsx
Goondrious: Yeah typescript + `useReducer` felt particularly awkward. This was more a result of my default habit of useReducer over useState. I don't think it's necessary here.

packages/ui/src/components/Tooltip.tsx
Goondrious: Yeup. I added this last for no particular reason other than tinkering.

Goondrious:
@andymatuschak base checks pass. Can sort out styling, configurability and anything else. I'm also around for the next few weeks if there are other tasks you'd like help with.

Issue #223, opened 2021-06-11
Add support to useKeyDown for React Native targets


andymatuschak:
In #218, @growthepie added some initial support for keyboard shortcuts in review sessions. We'll want that same support in the React Native apps, particularly the desktop app.

React Native does implement the `onkeydown` event on `View`, but our implementation of `useKeyDown` creates a _global_ handler by adding the event handler to `document`. That won't work on React Native, so we'll need to either create some coordination at the root level (maybe with a `Context` or something), or else use something like [react-native-keyevent](https://github.com/kevinejohn/react-native-keyevent) which offers more control.

PR #250, opened 2022-01-25
Add support for key commands in mac/iOS

kirkbyo:
Related https://github.com/andymatuschak/orbit/issues/223

## What?
Add support for key commands in macOS and iOS. 

## How?
I ended up integrating the iOS portion of [react-native-keyevent](https://github.com/kevinejohn/react-native-keyevent) since I couldn't find an acceptable solution using React Native API's (e.g view does not seem to support `onkeyevent`, only `TextInput` has access to key based callbacks AFAIK). I also only did the setup for iOS since I don't have any means to test the android platforms and I wasn't sure if there would be any utility there at this point anyways.

## How did I test this?
- Ran the macOS/Catalyst app pointed to dev
- Ran the macOS/Catalyst app pointed to prod and did my morning review using the key commands
- Ran the iPad app in the simulator
- Ran the web app pointed to dev to ensure no regressions

### DIFF
--- a/packages/app/ios/Orbit/AppDelegate.m
+++ b/packages/app/ios/Orbit/AppDelegate.m
@@ -15,6 +15,8 @@
 
 #import "ORDebugManager.h"
 
+#import <RNKeyEvent.h>
+
 #if 0
 //#if DEBUG && !TARGET_OS_MACCATALYST
 #import <FlipperKit/FlipperClient.h>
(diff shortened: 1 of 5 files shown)


andymatuschak (approved):
Great work, @kirkbyo! This all looks really sharp to me. One wondering: why did you choose to listen for key up events rather than key down on native platforms? Experimenting with a few macOS apps here, it seems that the norm for shortcut commands is to run them on key down. I'd probably want to follow that norm unless you had some reasoning otherwise?

kirkbyo:
Good call out @andymatuschak! I originally attempted to use the `onKeyDown` callback, however I run into this horrible crash:

Screen Shot 2022-01-27 at 6 21 26 PM

I'll go ahead an add a comment!

andymatuschak:
Got it. OK, we can investigate that later when it bothers us sufficiently. Thanks! :)

Issue #260, opened 2022-03-07
Proposal: Intermediate File Format to simplify note-sync syncing


kirkbyo:
Okay this is my proposal, for partially tackling https://github.com/andymatuschak/orbit/issues/220. @andymatuschak (and anyone else who might stumble upon this issue) any feedback is appreciated!

---

To tackle eventually tackle https://github.com/andymatuschak/orbit/issues/220 and https://github.com/andymatuschak/orbit/issues/246, I propose creating an intermediate file format to aid and encapsulate the syncing component of this problem.

**Why an intermediate format?**
- It enforces the seperation of concerns between syncing and parsing from various sources.
- Orbit (or extenernal actors) could define parsers for there source and produce an _Orbit Sync Document_ in which the sync library would translate to `store` events.
	- For example, Orbit could provide interpreters for Markdown and Anki as a source and generate sync documents.
- Stable Identifier agnostic. One of the current problems with the `note-sync` library is that it is coupled to Bear's export ID. Having an intermediate format, allows for other sources ([Obsidian Block Links](https://help.obsidian.md/How+to/Link+to+blocks#Link+to+blocks), Bear ID, Anki ID, etc.) to all provide the stable identifiers of there prompts. 
	- This can also help with two-sync problems down the road (sync from Markdown -> ingest into Orbit -> edit prompt in Orbit -> update original markdown note).
- Interpreters can be created in other programming languages. The file acts as the transport medium.

### Solution 1: Markdown as an Intermediate Format
One possible solution is to use markdown files augmented with [CommonMark generic directive syntax](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444) as the intermediate format. This might look something like:

```markdown
# Document Title

> Q. This is **Question 1**
> A. This is an answer to question 1
:orbit-prompt{prompt_id=d758904803cb}

> Q. This is **Question 2**
> A. This is an answer to question 2
:orbit-prompt{prompt_id=41b6d61c2aaa}

> This is a {cloze} prompt
:orbit-prompt{prompt_id=8656e0ee7d01}
```

This is just one example in how the format could be devised, if this solution is chosen, the format layout can be discussed further.

**Pro**:
- Common file format that can edited in various editors.
- Optimizes for developer readability.
- Leverages existing Markdown syntax to help with styling.

**Con**:
- Lacks expressiveness could lead to problems of ambiguity around markdown syntax (e.g https://github.com/andymatuschak/orbit/issues/186).
[...]

PR #266, opened 2022-04-25
Intermediate format to simplify syncing

kirkbyo:
Closes https://github.com/andymatuschak/orbit/issues/260

## What
This PR defines a new package called `Ingester` which outlines an intermediate file format to sync prompts to an `OrbitStore`.

## Why
Having this intermediate file format allows for the decoupling between source parsing and syncing. 

## What is the format?
```typescript
export type Ingestible = {
  sources: IngestibleSource[];
};

export interface IngestibleSource {
  // unique identifier representing a source which tasks might share, used for clustering; can be a URL or an arbitrary UUID
  identifier: IngestibleSourceIdentifier;
  // title that can be used when displaying the source.
  title: string;
  // prompts associated with the given source
  prompts: IngestiblePrompt[];
  // an optional URL to open when the user indicates they'd like to navigate to the provenance of the task
  url?: string;
  // an optional const that can be used to customize the visual appearance of the source during review
  colorPaletteName?: ColorPaletteName;
}
```

In this schema, producers (Anki, Bear, General Markdown, Biome) generate an "ingestible" file which are associated with an arbitrary number of sources. Prompts are associated with sources. Bear might produce a source per note, while Anki might produce a source per deck. A producer generates a stable identifier for each source. The ingester package is responsible for diffing between the prompts in the Orbit store, and the ones provided by the producer.

I believe this current format strikes a good balance between the "inferred" vs. "explicit" problem that I attempted to enumerate in the original issue. In this model, a producer does not need to output the absolute set of prompts. It can instead only produce sources that need to be synced. Sources which are not provided, are **not considered**. See `ingest.test.ts` test cases which I think capture the expected usage of this API.

## How has this been tested?
I am currently working on porting the `note-sync` package to use this code on https://github.com/kirkbyo/orbit/commits/bear-import. That work should help give some confidence in that this API is sound. That being said, I do not have a large corpus of markdown files with embedded prompts (and not a Bear user at that), so if you have any ideas on how I could test this further let me know.

### DIFF
--- a/packages/ingester/package.json
+++ b/packages/ingester/package.json
@@ -10,7 +10,7 @@
     "build": "tsc -b",
     "test": "jest --runInBand",
     "ingest": "yarn build; node --experimental-specifier-resolution=node dist/bin/run.js",
-    "generateSchema": "typescript-json-schema src/orbitAPI.ts ValidatableSpec -o src/orbitAPISchema.json --noExtraProps --required --ignoreErrors --strictNullChecks"
+    "generateSchema": "typescript-json-schema src/ingestible.ts Ingestible -o src/ingestible.json --noExtraProps --required --ignoreErrors --strictNullChecks"
   },
   "dependencies": {
     "@withorbit/core": "0.0.1",
--- a/packages/ingester/package.json
+++ b/packages/ingester/package.json
@@ -1,0 +1,29 @@
+{
+  "name": "@withorbit/ingester",
+  "version": "0.0.1",
+  "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "private": true,
+  "sideEffects": false,
+  "scripts": {
+    "build": "tsc -b",
+    "test": "jest --runInBand",
+    "ingest": "yarn build; node --experimental-specifier-resolution=node dist/bin/run.js",
+    "generateSchema": "typescript-json-schema src/orbitAPI.ts ValidatableSpec -o src/orbitAPISchema.json --noExtraProps --required --ignoreErrors --strictNullChecks"
+  },
+  "dependencies": {
+    "@withorbit/core": "0.0.1",
+    "@withorbit/store-shared": "0.0.1",
+    "@withorbit/store-fs": "0.0.1",
+    "ajv": "^8.6.2"
+  },
+  "devDependencies": {
+    "@types/jest": "^26.0.24",
+    "@types/node": "^14.14.7",
+    "babel-jest": "^27.0.6",
+    "jest": "^27.0.6",
+    "typescript": "^4.2.4",
+    "typescript-json-schema": "^0.50.0"
+  }
+}
--- a/packages/ingester/package.json
+++ b/packages/ingester/package.json
@@ -14,8 +14,8 @@
   },
   "dependencies": {
     "@withorbit/core": "0.0.1",
-    "@withorbit/store-shared": "0.0.1",
     "@withorbit/store-fs": "0.0.1",
+    "@withorbit/store-shared": "0.0.1",
     "ajv": "^8.6.2"
   },
   "devDependencies": {
--- a/packages/ingester/src/ingest.ts
+++ b/packages/ingester/src/ingest.ts
@@ -1,0 +1,41 @@
+import { ColorPaletteName, EntityType, Task } from "@withorbit/core";
+import { DatabaseEntityQuery, OrbitStore } from "@withorbit/store-shared";
+
+/**
+ * @TJS-type string
+ * @TJS-pattern ^[0-9a-zA-Z_\-]{22}$
+ */
+type IngestibleSourceIdentifier = string & {
+  __sourceIDOpaqueType: never;
+};
+
+export interface IngestibleSource {
+  // unique identifier representing a source which tasks might share, used for clustering; can be a URL or an arbitrary UUID
+  identifier: IngestibleSourceIdentifier;
+  // title that can be used when displaying the source.
+  title: string;
+  // prompts associated with the given source
+  prompts: IngestiblePrompt[];
+  // an optional URL to open when the user indicates they'd like to navigate to the provenance of the task
+  url?: string;
+  // an optional const that can be used to customize the visual appearance of the source during review
+  colorPaletteName?: ColorPaletteName;
+}
+
+type IngestiblePrompt = IngestibleQAPrompt;
+
+export interface IngestibleQAPrompt {
+  body: { text: string };
+  answer: { text: string };
+}
+
+export async function ingestSources(
+  sources: IngestibleSource[],
+  store: OrbitStore,
+) {
+  const query: DatabaseEntityQuery<Task> = {
+    entityType: EntityType.Task,
+  };
+  const entities = await store.database.listEntities(query);
+  console.log(entities);
+}
--- a/packages/ingester/src/ingest.ts
+++ b/packages/ingester/src/ingest.ts
@@ -17,13 +17,18 @@ import {
 } from "./ingestible";
 
 type IngestOptions = {
-  ingestDateMilis: number;
+  ingestDateMillis: number;
 };
 const DEFAULT_OPTIONS = (): IngestOptions => ({
-  ingestDateMilis: Date.now(),
+  ingestDateMillis: Date.now(),
 });
 
-export const INGEST_ITEM_IDENTIFIER_KEY = "ingest_item_identifier" as const;
+export const INGEST_ITEM_IDENTIFIER_KEY = "ingestSourceIdentifier" as const;
+export const MissingItemIdentifierError = new Error(
+  "existing item with matching source identifier does not contain item identifier",
+);
+export const DuplicateItemIdentifierError = (identifier: string) =>
+  new Error(`item identifier '${identifier}' is not unique`);
 
 // TODO: handle moves across sources
 // TODO: handle provenance updates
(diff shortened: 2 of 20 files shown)


kirkbyo:
@andymatuschak This PR is in a good state in case you wanted to give it a general first pass review for any major pitfalls that I might have overlooked!

Remaining todo's on my end:

- [ ] Support cloze prompts
- [ ] Support individual prompts having an `openURL`
- [ ] Finish migrating the `note-sync` package to use this package in a separate branch. This will help ensure that the current implementation is sound e2e _(note: the actual migration work will be reviewed in a separate PR)_.

packages/store-fs/src/sqlite.ts
```diff
@@ -44,10 +44,13 @@ export class SQLDatabaseBackend implements DatabaseBackend {
 
   static attachmentURLProtocol = "x-orbit-sqlattachment";
 
-  constructor(path: string) {
+  constructor(path: string, options?: { enableDebugLogs?: boolean }) {
```
kirkbyo, line 47: these logs were becoming quite noisy in the unit tests so I created a flag to disable them. Happy to revert these changes and just stub out `console.log` in my tests if you would prefer.

The more SOLID approach would to pass in a `Logger` interface, but that felt a bit heavy for this case.

packages/ingester/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@withorbit/ingester",
```
kirkbyo, line 2: Naming is still up in the air, let me know if you feel if something else might be more apt!

I also didn't license this package, let me know if you would like me to port over any of the existing licenses into this package.

packages/ingester/src/ingestible.ts
```diff
@@ -0,0 +1,34 @@
...
+  __sourceIDOpaqueType: never;
+};
+
+export interface IngestibleSource {
+  // unique identifier representing a source which tasks might share, used for clustering; can be a URL or an arbitrary UUID
+  identifier: IngestibleSourceIdentifier;
+  // title that can be used when displaying the source.
+  title: string;
+  // prompts associated with the given source
+  prompts: IngestiblePrompt[];
```
andymatuschak, line 21: Two somewhat invasive suggestions; feel free to take or leave.

**Prompt -> TaskSpec**
I think it might be more parsimonious to replace `prompts` with `taskSpecs` here? I think `TaskSpec` is the general form of what you're pointing to here. It would let you avoid redefining QA prompts below and mapping between tasks and prompts. And once we define non-memory tasks (which I do intend to do!) I expect to want to ingest them from Markdown in a similar way.

A very simple task type I expect to add would simply display some content, and then provide some controls which let you adjust the subsequent interval. I might use this to show myself visual inspiration, or for tasks like "sing a minor sixth upward", etc. I can imagine wanting to ingest these from my Markdown notes by tagging any block with `#orbitTask`.

Feel free to YAGNI me.

**Provenance**
For that matter, the other properties seem mostly duplicative of `TaskProvenance`. Could this whole thing become?:

```
export interface IngestibleSource {
   // unique identifier representing a source which tasks might share, used for clustering; can be a URL or an arbitrary UUID
   identifier: IngestibleSourceIdentifier;

   specs: TaskSpec[];
   provenance: TaskProvenance;
}

File: packages/ingester/src/ingest.ts [hunk 4]
```diff
@@ -0,0 +1,154 @@
...
+
+  for (const source of sources) {
+    if (groupedEntities[source.identifier]) {
+      // determine which prompts are new and which have been deleted
+      const existingPrompts = [...groupedEntities[source.identifier]];
+
+      for (const prompt of source.prompts) {
+        // NOTE: this `findIndex` is a bit expensive since it scans of each iteration.
+        // If this becomes a bottleneck we should precompute a hash for each task prompt.
+        const existingPromptIndex = existingPrompts.findIndex(
```
andymatuschak, line 43: This suggestion complicates your design here, but I figure it's worth discussing:

In the present design, ingestion sources surface a bunch of prompt content (i.e. front/back strings). The shared ingestion machinery implements a notion of "identity" on top of this by matching that verbatim content with existing prompt content.

This is expensive, but it might also be more limiting than it needs to be. I implemented that policy for the Markdown importer because that format has no way of tracking identity over time. But Biome could very reasonably make a little ID for each prompt you create in its UI! We could be using those IDs to maintain a prompt's identity even when its text is modified.

Concretely, an alternative would be to have `IngestionPrompt` include an arbitrary identifier, provided by the source. For Markdown prompts, that'd just be a hash of the content; for other sources it might be some stable identifier. We'd store that in a metadata key on the `Task`. Then we could replace this double-loop with set arithmetic.

(Feel free to YAGNI me of course)

andymatuschak (commented):
Exciting!! I made a few high-level suggestions for discussion; I didn't finish looking through in detail because things will probably change if you choose to follow any of them. But feel free to reject all these and continue directly—I don't want to complicate. Just let me know what you'd prefer!

kirkbyo (commented):
Thanks for the first pass! This PR should be good to now.

A couple things to tackle in follow up PRs:
- I have opened https://github.com/andymatuschak/orbit/pull/268 which migrates the note-sync to use this work, and is also how I tested this code e2e to ensure it is working. That being said, I do not have a corpus of markdown notes with prompts so I can't assess the performance of this work.
- Now that task items have an identifier associated that is stored in prompt metadata, we will need a migration script to allow existing users of the `note-sync` package to migrate to the interpreter above.
- Provenance updates and moves are currently not tracked in this work. Implementing this behaviour is relatively straightforward with the current setup, but I didn't want to complicate this review any further.

packages/ingester/src/ingest.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+  return mapping;
+}
+
+function groupEntitiesBySourceIdentifiers(entities: Task<TaskContent>[]) {
+  // [01/05/22]: our current versin of typescript does not allow
+  // `IngestibleItemIdentifier` to be the key of an object. Newer
+  // releases of typescript support this behavior.
+  // https://github.com/microsoft/TypeScript/pull/44512
+  const mapping: Record<string, Task<TaskContent>> = {};
```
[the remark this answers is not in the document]
andymatuschak: Unfortunate, but sounds like the right call.

packages/ingester/package.json
andymatuschak: `ingester` is fine for now. It'd be great if you could copy the Apache `LICENSE` from one of the other library packages into the root here.

packages/ingester/src/ingest.ts
```diff
@@ -0,0 +1,160 @@
...
+) {
+  const events: Event[] = [];
+
+  // TODO: create a new query or extend the entity query such that
+  // we can filter on only sources that would have used this ingester
+  const query: DatabaseEntityQuery<Task> = {
+    entityType: EntityType.Task,
+  };
+  const entities = await store.database.listEntities(query);
+  const groupedEntities = groupEntitiesByProvenanceIdentifiers(entities);
```
andymatuschak, line 43: Reading this, I noticed that I had a lot of trouble following these different maps: what's being mapped to what? One thing I liked about Cocoa was its very consistent naming conventions—meant less thinking while coding, since names were largely formulaic, but also helped legibility. Maps/dictionaries/records used the convention `valuesByKey`, so the three maps here might be named: `existingEntitiesByProvenanceIdentifiers` (I added "existing" here because that distinction is important in this routine), `existingEntitiesByItemIdentifiers`, and `ingestibleItemsByItemIdentifiers`.

No obligation to change this, just thought you might find it interesting.

packages/ingester/src/ingest.ts
```diff
@@ -0,0 +1,160 @@
...
+  const timeMillis = opts.ingestDateMilis;
+
+  for (const source of sources) {
+    if (groupedEntities[source.identifier]) {
+      // determine which prompts are new and which have been deleted
+      const existingPrompts = mapEntitiesByItemIdentifiers(
```
andymatuschak, line 49: Prompts? Tasks? Entities?

packages/ingester/src/ingest.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (mapping[provenance.identifier]) {
+      mapping[provenance.identifier].push(entity);
+    } else {
+      mapping[provenance.identifier] = [entity];
+    }
+  }
+  return mapping;
+}
+
+function mapEntitiesByItemIdentifiers(entities: Task<TaskContent>[]) {
```
andymatuschak, line 99: Thanks for renaming. :)

packages/ingester/src/validateIngestible.test.ts
```diff
@@ -0,0 +1,35 @@
+import { createIngestibleValidator } from "./validateIngestible";
+
+it("accepts valid schema", () => {
+  const validator = createIngestibleValidator({
+    mutateWithDefaultValues: true,
+  });
+  const { isValid, errors } = validator.validate({
+    sources: [
+      {
+        identifier: "aaaaaaaaaaaaaaaaaaaaaa",
```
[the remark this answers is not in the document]
andymatuschak: Recapping our phone discussion: no need to validate these external IDs; we don't/can't know what scheme they're using.

andymatuschak (approved):
Looking great! Some comments throughout, but feel free to take/leave all. This is ready when you're ready.

packages/ingester/src/ingest.ts
andymatuschak: Ah, the naming is complicated slightly by the fact that the first map is actually of _groups_; if this were Cocoa I'd probably name it `existingEntityGroupsByProvenanceIdentifiers` (of course, Cocoa likes long variable names, and that's perhaps less idiomatic in JS culture)

packages/ingester/src/ingest.ts
```diff
@@ -0,0 +1,160 @@
...
+        createIngestTaskForSource(source, timeMillis),
+      );
+      events.push(...ingestEvents);
+    }
+  }
+  return events;
+}
+
+function groupEntitiesByProvenanceIdentifiers(entities: Task<TaskContent>[]) {
+  const mapping: Record<TaskProvenance["identifier"], Task<TaskContent>[]> = {};
```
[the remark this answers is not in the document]
kirkbyo: Oh TIL! I didn't realize JS had a map construct. That is great!

kirkbyo:
Thanks for the review @andymatuschak! I believe I addressed all the comments and also added two new error conditions: duplicate item identifiers and missing item identifier in the metadata. I am going to go ahead and merge this so that the other PR's can be rebased 🚀

Issue #220, opened 2021-06-04
Sync notes from 


jessmartin:
I've been looking into the problem of how to sync SRS prompts from note-taking systems with Orbit. 

The goal is to have a CLI tool that:
1. Takes as input an arbitrary set of notes, most likely a folder full of markdown-formatted text files
2. Parses those notes for cloze deletion and Q and A style prompts
3. Syncs those prompts to Orbit

The `note-sync` library already accomplishes most of this, but it has a few shortcomings:
- It depends on a larger set of libraries that Andy wrote to solve a more general problem (`computer-supported-thinking`, `spaced-everything`, and `incremental-thinking`), which makes the code more complex than necessary.
- Markdown files are expected to have a certain format consistent with being exported from Bear.
- Prompts are being cached locally, which feels unnecessary, both from the perspective of performance and code complexity. (see note below about necessity of caching)

After spending a few hours reviewing the code and testing out the `note-sync` package, I have a few thoughts/opinions:
- It seems like it would be simpler to do a complete re-write. The one exception is the parsing code in `incremental-thinking` that parses `qaPrompts` and `clozePrompts`. That code is well-tested and has been in use by Andy for a while.
- I don't think this package should support syncing prompts from markdown notes -> Anki. Given this package's scope is Orbit, I think it should focus on sync with Orbit only. Maybe Anki and Orbit should be able to sync with one another, so that Anki could be used as a review interface? If so, that feels like a separate concern. It looks like there is a package called `anki-import` that at least handles one-way importing.
- I don't think this package should do any local caching. Based on the discussion of [Idempotency and Identity](https://github.com/andymatuschak/orbit/tree/master/packages/note-sync#idempotency-and-identity), we should be able to simply compare hashes of the prompts to know whether a prompt is new or not. We could debate whether the Orbit API should allow a duplicate prompt to be created, but at worst, we only need to grab the hashes for all existing prompts from the Orbit API and then do some hash comparisons. I admit that I don't fully understand the caching code yet, so maybe I'm missing something!

## Thoughts on Provenance

How should this library handle provenance? Broadly, I have questions about how Orbit thinks about provenance, but scoping my questions to this library, it appears that the current implementation is caching provenance information locally, but not syncing that provenance information to the Orbit API.

The current implementation depends on a Bear Note ID at the bottom of the markdown file to determine provenance, which is obviously undesirable as notes could be exported from a variety of different note-taking systems.

[...]

PR #268, opened 2022-05-01
Replicate behaviour of note-sync package to use ingester

kirkbyo:
Relates https://github.com/andymatuschak/orbit/issues/260, https://github.com/andymatuschak/orbit/issues/220

**What?**
Migrate the `note-sync` package to use the new ingester schema. 

**How has this been tested?**
I have used the bear markdown export script to generate a small set of test files to `interpret` against.

### DIFF
--- a/packages/interpreter/src/bin/run.ts
+++ b/packages/interpreter/src/bin/run.ts
@@ -1,0 +1,62 @@
+import { Ingestible } from "@withorbit/ingester";
+import fs from "fs";
+import path from "path";
+import { BearInterpreter } from "../interpreters";
+import { InterpretableFile, Interpreter } from "../interpreter";
+
+async function run(
+  noteDirectory: string,
+  interpreter: Interpreter,
+): Promise<Ingestible> {
+  // read and parse each file in the directory
+  const fileNames = await listNoteFiles(noteDirectory);
+  const files = await Promise.all(
+    fileNames.map(async (fileName): Promise<InterpretableFile> => {
+      const filePath = path.join(noteDirectory, fileName);
+      const content = await readFile(filePath);
+      return {
+        name: fileName,
+        path: filePath,
+        content,
+      };
+    }),
+  );
+  return interpreter.interpret(files);
+}
+
+async function listNoteFiles(folderPath: string): Promise<string[]> {
+  const noteDirectoryEntries = await fs.promises.readdir(folderPath, {
+    withFileTypes: true,
+  });
+  return noteDirectoryEntries
+    .filter(
+      (entry) =>
+        entry.isFile() &&
+        !entry.name.startsWith(".") &&
+        entry.name.endsWith(".md"),
+    )
+    .map((entry) => entry.name);
+}
+
+async function readFile(path: string): Promise<string> {
+  const buffer = await fs.promises.readFile(path);
+  return buffer.toString("utf-8");
+}
+
+(async () => {
+  // ensure that the arguments are defined
+  const noteDirectory = process.argv[2];
+  const outJsonFilePath = process.argv[3];
+  if (!noteDirectory || !outJsonFilePath) {
+    console.error("Usage: yarn ingest /path/to/notes /path/to/json-file");
+    process.exit(1);
+  }
+
+  const interpreter = new BearInterpreter();
+  const ingestible = await run(noteDirectory, interpreter);
+  await fs.promises.writeFile(outJsonFilePath, JSON.stringify(ingestible));
+})()
+  .then(() => {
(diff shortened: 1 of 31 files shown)


packages/interpreter/src/interpreters/bear/plugins/backlinksPlugin.test.ts
```diff
@@ -0,0 +1,102 @@
+import mdast from "mdast";
```
kirkbyo, line 1: all files in the `packages/interpreter/src/interpreters/bear/plugins` directory have been directly ported from the vendored `incremental-thinking` package. I thought moving these over might more sense then trying to vendor the package here again.

packages/interpreter/src/bin/run.ts
```diff
@@ -0,0 +1,63 @@
...
+  const buffer = await fs.promises.readFile(path);
+  return buffer.toString("utf-8");
+}
+
+(async () => {
+  // ensure that the arguments are defined
+  const noteDirectory = process.argv[2];
+  const outJsonFilePath = process.argv[3];
+  if (!noteDirectory || !outJsonFilePath) {
+    console.error("Usage: yarn ingest /path/to/notes /path/to/json-file");
```
andymatuschak, line 52: `interpret` now, right?

packages/interpreter/src/interpreters/bear/BearInterpreter.ts
```diff
@@ -0,0 +1,92 @@
...
+      }),
+    );
+    const sources = nullableSources.filter(
+      (source) => source !== null,
+    ) as IngestibleSource[];
+    return { sources };
+  }
+}
+
+function convertInterpreterPromptToIngestable(prompt: Prompt): TaskSpec {
```
andymatuschak, line 60: sp: Ingestible

packages/interpreter/src/interpreters/bear/BearInterpreter.ts
```diff
@@ -0,0 +1,92 @@
...
+      files.map(async (file): Promise<IngestibleSource | null> => {
+        const root = await parseMarkdown(file.content);
+        const bearId = getStableBearID(root);
+
+        if (bearId) {
+          const prompts = findAllPrompts(root);
+          const noteTitle = getNoteTitle(root);
+
+          return {
+            identifier: (bearId?.id ?? file.path) as IngestibleSourceIdentifier,
```
andymatuschak, line 37: `bearId` is guaranteed non-null here, right?

packages/interpreter/src/interpreters/bear/plugins/backlinksPlugin.test.ts
andymatuschak: Yes, I think that's the right call. I don't think that other package is going anywhere.

packages/interpreter/src/interpreters/bear/plugins/noteLinkProcessorPlugin.ts
```diff
@@ -0,0 +1,48 @@
...
+        targetNoteName: wikiLinkNode.value,
+        targetNoteDisplayName: wikiLinkNode.data.alias,
+      } as NoteLinkNode;
+    } else {
+      return visitee;
+    }
+  });
+}
+
+export default function noteLinkPlugin(this: unified.Processor) {
```
andymatuschak, line 30: I don't think the interpreter actually uses this plugin—may be able to just drop this file altogether.

andymatuschak (approved):
Looking great. Thank you for your diligently keeping all these tests in place as you migrated the parser over.

Just a few tiny things inline. One high-level request: would you copy the Apache `LICENSE` file from one of the other library packages into the root of this one?

kirkbyo (commented):
Thanks for the review! I rebased it against master and applied the feedback. I also went ahead and tested it against the markdown notes you provided and everything seems to be working correctly 😄  

Some stats:
- It took 3.54s to parse the notes you provided and generate an ingestible file. I suspect this could be improved in the future by only reading and parsing files that we know to have been modified since our last sync.
- It took 1.3s to sync the ingestible prompts to a copy of my orbit store. Very snappy.

Let me know if anything else stands out, otherwise feel free to merge 😄

packages/interpreter/src/interpreters/bear/plugins/noteLinkProcessorPlugin.ts
andymatuschak: We do, for a simple silly reason: to ignore embedded prompts which are included in the backlinks sections of notes!

PR #269, opened 2022-05-09
Add SiriKit intent to trigger ingester

kirkbyo:
Now that https://github.com/andymatuschak/orbit/pull/266 has landed, its API can be exposed to Apple's SiriKit Intents. This will allow apps like Biome to sync their prompts directly to Orbit using the Shortcuts app.

**How has this been tested?**
I created a SiriKit intent which generates an Orbit ingestible based on Biome prompts. This intent will feed its file to the Orbit "Ingester" intent. 

Screen Shot 2022-05-08 at 5 21 50 PM


packages/app/ios/Orbit/IngestEventEmitter.h
```diff
@@ -0,0 +1,30 @@
+//
+//  EventEmitter.h
+//  Orbit
+//
+//  Created by Ozzie Kirkby on 2022-05-07.
+//
+
+#ifndef EventEmitter_h
```
kirkbyo, line 8: Note: this is my first time writing any sort of meaningful amount of Objective-C so let me know if anything stands out!

packages/app/ios/Orbit/IngestEventEmitter.m
```diff
@@ -0,0 +1,78 @@
...
+}
+
+RCT_EXPORT_METHOD(completedIngestion:(BOOL *)success) {
+  if (success) {
+    self.intentState = kSuccess;
+  } else {
+    self.intentState = kFailure;
+  }
+  dispatch_semaphore_signal([self semaphore]);
+}
```
kirkbyo, line 75: Let me know if you know of a better way to structure this, but the problem is this: `emitIngestEvent` needs to return if an ingest event was successful (i.e was it able to ingest correctly), however React Native's `sendEventWithName` happens async. To get around this, I setup this blocking semaphore which is resolved when the static  `completedIngestion` is called. This of course a bit fragile to `completedIngestion` always being called, but it seemed like the most reasonable work around I could find.

packages/app/ios/Orbit/Intents.intentdefinition
```diff
@@ -0,0 +1,154 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kirkbyo, line 1: Reading this XML doesn't make it clear how this is setup, here is the UI provided by Xcode.
Screen Shot 2022-05-08 at 5 46 31 PM

Screen Shot 2022-05-08 at 5 46 39 PM

packages/app/src/model2/orbitStoreFactory.ts
```diff
@@ -6,3 +6,7 @@ export async function createOrbitStore(
 ): Promise<OrbitStore> {
   return new OrbitStoreFS(databaseName);
 }
+
+export async function createDefaultOrbitStore(): Promise<OrbitStore> {
```
kirkbyo, line 10: moved this to be a helper function so that the ingester handler can init the store outside the React lifeycle.

packages/app/ios/Orbit/IngestEventEmitter.h
```diff
@@ -0,0 +1,30 @@
...
+#define EventEmitter_h
+
+#import <React/RCTBridgeModule.h>
+#import <React/RCTEventEmitter.h>
+
+typedef enum {
+    kSuccess,
+    kFailure,
+    kWaiting
+} IntentState;
```
andymatuschak, line 18: You asked for Obj-C style notes, so you'll get them! Feel free to ignore these…

Idiomatic ObjC enum naming would be:

```
typedef enum {
  IntentStateSuccess,
  IntentStateFailure,
  IntentStateWaiting
} IntentState
```

This is because Obj-C doesn't have the TS syntax of `EnumName.Value`—it's just `Value`. The prefix makes for nice autocomplete and readability. See e.g. `UIImageResizingMode`.

packages/app/ios/Orbit/IngestEventEmitter.h
```diff
@@ -0,0 +1,30 @@
...
+
+@interface IngestEventEmitter : RCTEventEmitter <RCTBridgeModule>
+
+@property(atomic,assign) bool hasListeners;
```
andymatuschak, line 22: There isn't a consistent style for spacing in this syntax, but I'd suggest making 24 look like 22/23.

packages/app/ios/Orbit/IngestEventEmitter.h
```diff
@@ -0,0 +1,30 @@
...
+@property(atomic,assign) IntentState intentState;
+@property (atomic,assign) dispatch_semaphore_t semaphore;
+
+- (bool) emitIngestEvent:(NSString *) fileJSON;
```
andymatuschak, line 26: No space after the types:

```
- (bool)emitIngestEvent:(NSString *)fileJSON

File: packages/app/ios/Orbit/IngestEventEmitter.m [hunk 8]
```diff
@@ -0,0 +1,78 @@
...
+
+#import "IngestEventEmitter.h"
+#import <libkern/OSAtomic.h>
+#import <stdatomic.h>
+
+@implementation IngestEventEmitter
+
+@synthesize semaphore;
+
+- (instancetype)init {
```
andymatuschak, line 18: Shouldn't be necessary, since you're not doing anything other than calling super.

packages/app/ios/Orbit/IngestEventEmitter.m
```diff
@@ -0,0 +1,78 @@
...
+  self = [super init];
+  return self;
+}
+
++ (BOOL)requiresMainQueueSetup {
+    return NO;
+}
+
+-(void)startObserving {
```
andymatuschak, line 27: `- (void)startObserving`

packages/app/ios/Orbit/IngestEventEmitter.m
```diff
@@ -0,0 +1,78 @@
...
+    self.hasListeners = YES;
+}
+
+// Will be called when this module's last listener is removed, or on dealloc.
+-(void)stopObserving {
```
andymatuschak, line 32: `- (void)stopObserving`

packages/app/ios/Orbit/IngestEventEmitter.m
```diff
@@ -0,0 +1,78 @@
...
+        sharedInstance = [super allocWithZone:zone];
+    });
+    return sharedInstance;
+}
+
+- (NSArray<NSString *> *)supportedEvents {
+    return @[@"onIngestEvent"];
+}
+
+- (bool) emitIngestEvent:(NSString *) fileJSON {
```
andymatuschak, line 51: `- (bool)emitIngestEvent:(NSString *)fileJSON`

andymatuschak: For [weird historical reasons](https://stackoverflow.com/questions/541289/objective-c-bool-vs-bool), all `bool` should be `BOOL` in Objective-C—they're encoded slightly differently, and in a way Foundation expects.

packages/app/ios/Orbit/IngestEventEmitter.m
andymatuschak: Oof, yikes. Yeah, I've run into this kind of problem before, and I've had to do this kind of thing.

Happily, in this case, it looks like you may be able to dodge the problem. AFAICT `emitIngestEvent` is only used in `handleIngest:completion:`, but that method's already set up to be async, right? Can you call the completion handler from there after a response comes back?

packages/app/ios/Orbit/IngestEventHandler.h
```diff
@@ -0,0 +1,16 @@
+//
+//  IngestEventHandler.h
+//  Orbit
+//
+//  Created by Ozzie Kirkby on 2022-05-08.
+//
+
+#ifndef IngestEventHandler_h
+#define IngestEventHandler_h
+#import "IngestIntent.h"
```
andymatuschak, line 10: Looks like this file didn't get added!

packages/app/ios/Orbit/IngestEventHandler.m
```diff
@@ -0,0 +1,35 @@
...
+
+#import <Foundation/Foundation.h>
+#import "IngestEventHandler.h"
+#import "IngestIntent.h"
+#import "IngestEventEmitter.h"
+
+@implementation IngestEventHandler
+
+- (void)handleIngest:(IngestIntent *)intent completion:(void (^)(IngestIntentResponse *response))completion NS_SWIFT_NAME(handle(intent:completion:)) {
+  IngestIntent *ingestIntent = ((IngestIntent *) intent);
```
andymatuschak, line 16: `= (IngestIntent *)intent;`

packages/app/ios/Orbit/IngestEventHandler.m
```diff
@@ -0,0 +1,35 @@
...
+  NSString *data = [[NSString alloc] initWithData: ingestIntent.file.data encoding: NSUTF8StringEncoding];
```
andymatuschak, line 17: No spaces after colons

andymatuschak (changes requested):
Exciting!! I couldn't quite finish the review because it seems like a file was missing.

Sorry for all the style notes… I wish there were something like `prettier` for Obj-C. This is a helpful style guide: https://google.github.io/styleguide/objcguide.html

kirkbyo:
Thank you for the review and all the style notes. You can tell the code I wrote vs. the code I copied; my Swift/TS biases were leaking through 😄. Also, good catch on the completion handler, that completely escaped me. Much cleaner now.

Regarding the missing file, that file is automatically generated by SiriKit on build. It generates a set of typings based on the Intent definition file. I ended up adding a comment above the import and renaming the intent to better reflect this.

andymatuschak:
Ah, very interesting that it’s auto generated! That’s an unusual API design approach for Apple. I guess it probably fits better with Swift aesthetics than Obj-C, which emphasizes runtime reflection.

Thank you for this! Please merge when ready.

andymatuschak (approved):

kirkbyo:
Okay finally good to land this. Finished implementing some truly stable identifiers in Biome and confirmed everything is working e2e!

PR #275, opened 2022-06-18
web - add missing function

maciejmatu:
While trying to set up the project, I ran into a small issue. It seems like there was a function missing for `orbitStoreFactory.web.ts`. I also moved the `store-shared` import to be the first one in both files 😅 

The function I added is the same for both files, so maybe it would make sense to create a 3rd intermediary file, that would contain `createDefaultOrbitStore`, though it's down to your preference.

image

### DIFF
--- a/packages/app/src/model2/orbitStoreFactory.ts
+++ b/packages/app/src/model2/orbitStoreFactory.ts
@@ -1,5 +1,5 @@
-import OrbitStoreFS from "@withorbit/store-fs";
 import { OrbitStore } from "@withorbit/store-shared";
+import OrbitStoreFS from "@withorbit/store-fs";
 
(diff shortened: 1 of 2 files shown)


kirkbyo:
Ah thanks for the fix @maciejmatu! This was due to my change in https://github.com/andymatuschak/orbit/pull/266. I was able to reproduce and confirm this fix worked locally for me as well. 

cc @andymatuschak since I do not have any context on why need to duplicate across these two files.

andymatuschak:
Ah, a bit of magic: when present, `.web.ts` variants will be used over `.ts` when building for web. Likewise `.ios.ts`, `.android.ts`, `.native.ts` (i.e. any React Native target, but not Node or web). Thank you!

Issue #54, opened 2020-06-02
note-sync: implement support for “move” operations


andymatuschak:
Right now, when prompts are moved between notes, they're treated as successive deletions / additions. There's support in spaced-everything for "moves" but I haven't hooked it up for Orbit yet. A "move" in Orbit would mean a change in provenance metadata.

PR #276, opened 2022-06-20
Ingester: add support for moving items across sources

kirkbyo:
Related: https://github.com/andymatuschak/orbit/issues/54

Right now if you move a prompt to a different source, it will result in the prompt to be marked as deleted in the old source, but then re-added in the new source. Instead, if the same item identifier is associated with a delete and ingest event within the same sync than we can just generate a provenance update instead.

### DIFF
--- a/packages/ingester/src/ingest.ts
+++ b/packages/ingester/src/ingest.ts
@@ -5,9 +5,11 @@ import {
   generateUniqueID,
   Task,
   TaskContent,
+  TaskID,
   TaskIngestEvent,
   TaskProvenance,
   TaskUpdateDeletedEvent,
+  TaskUpdateProvenanceEvent,
 } from "@withorbit/core";
 import { DatabaseEntityQuery, OrbitStore } from "@withorbit/store-shared";
 import {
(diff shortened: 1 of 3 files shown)


packages/ingester/src/ingest.ts
```diff
@@ -62,30 +67,58 @@ export async function ingestSources(
...
         if (!existingEntitiesByItemIdentifiers.get(key)) {
           // task is new
-          events.push(
+          ingestEvents.set(
+            key,
             createIngestTaskForSource(source, timeMillis)(potentialNewItem),
           );
+        } else {
+          // task exists
+          // TODO: we should make sure the content is up to date
```
kirkbyo, line 76: @andymatuschak I just realized that I haven't implemented this case yet. Does Orbit have an event for updating a `Task`'s content yet? Would you just issue a new `TaskIngestEvent` and associate it with the same entity id? Happy to dig into this further; just want to make sure this isn't a solved problem already.

yarn.lock
```diff
@@ -11851,30 +11851,10 @@ caniuse-api@^3.0.0:
...
-caniuse-lite@^1.0.30001219:
-  version "1.0.30001245"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001245.tgz#45b941bbd833cb0fa53861ff2bae746b3c6ca5d4"
-  integrity sha512-768fM9j1PKXpOCKws6eTo3RHmvTUsG9UrpT4WoREFeZgJBTi4/X9g565azS/rVUGtqb8nt7FjLeF5u4kukERnA==
-
-caniuse-lite@^1.0.30001332:
-  version "1.0.30001332"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001332.tgz#39476d3aa8d83ea76359c70302eafdd4a1d727dd"
-  integrity sha512-10T30NYOEQtN6C11YGg411yebhvpnC6Z102+B95eAsN0oB6KUs01ivE8u+G6FMIRtIrVlYXhL+LUwQ3/hXwDWw==
+caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001125, caniuse-lite@^1.0.30001135, caniuse-lite@^1.0.30001157, caniuse-lite@^1.0.30001181, caniuse-lite@^1.0.30001219, caniuse-lite@^1.0.30001332:
```
kirkbyo, line 11854: got tired of seeing this warning every time I ran the tests. Happy to break this into a separate PR if you would prefer.

Screen Shot 2022-06-20 at 4 38 12 PM

packages/ingester/src/ingest.ts
andymatuschak: No, not yet! I was thinking something like a "spec update" event, probably with a sparse spec.

andymatuschak (approved):
Looks good! Please merge when ready.

PR #281, opened 2022-09-07
generalize bear interpreter to work for any markdown file

kirkbyo:


andymatuschak (approved):
This looks great! Nice how light a change this was.

Lightweight wondering: do you think it would it be better to use relative paths for non-Bear Markdown file identifiers? With the current arrangement, if you were to move your notes to a different folder on your computer, all those sources would have different identifiers. I'm not sure if that's a real problem, since the prompt histories would follow the actual text hashes of the prompts.

PR #286, opened 2022-11-07
install firebase-tools in only devDependencies

Goondrious:
Got an install error related to this, s removed from `packages/backend/package.json > dependencies` and bumped the version in devDependencies


andymatuschak:
Cool, thanks—you're right that we shouldn't need `firebase-tools` outside development.

